PR #75, opened 2025-05-09
Disable some Compose features if Docker's extension is installed

rcjsuen:
Docker's extension bundles a language server that provides some Compose features that overlap with the ones from the Container Tools extension. Red Hat's YAML extension also does something similar which was previously accounted for. The code has now been changed to disable select Compose features if Docker's extension or Red Hat's YAML extension is installed.

To test this, one should install [specific builds of Docker's extension](https://github.com/docker/vscode-extension/actions/runs/14888843267) that are being generated by https://github.com/docker/vscode-extension/pull/108.

### DIFF
--- a/src/utils/AlternateYamlLanguageServiceClientFeature.ts
+++ b/src/utils/AlternateYamlLanguageServiceClientFeature.ts
@@ -24,20 +24,28 @@ export class AlternateYamlLanguageServiceClientFeature implements StaticFeature,
         };
     }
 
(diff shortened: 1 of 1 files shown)


rcjsuen:
@bwateratmsft I've converted this to a draft for now. Will switch it back after we've figured out the coordination of things to ensure things are generally smooth and non-disruptive.

bwateratmsft:
Sounds good. Code looks great, thanks @rcjsuen!

bwateratmsft (approved):

PR #82, opened 2025-05-19
Add Compose Down Selected Services Command

bernardoduarte:
Add a new command for Compose, basically making a copy from current Compose Up - Selected Services for Compose Down command.

I've written it because I'm usually turning on and off some of my containers inside my docker-compose files, so to make my life easier, and perhaps others, I thought that it would be a nice addition to this new VSCode extension.


bernardoduarte:

RainbowScientist5 (approved):

danegsta (approved):

bwateratmsft:
This is now released in version 2.2.0 of the Container Tools extension.

PR #204, opened 2025-08-22
Suppress C# Dev Kit task prompt

### LINKED ISSUE #162: When C# Dev Kit extension is installed, a warning is shown that the user is likely to stumble over
Image

If the user selects "Yes", then our containerized debugging will not work.

bwateratmsft:
Fixes #162

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -71,7 +71,7 @@ const config = {
         // Copy some needed resource files from external sources
         new CopyPlugin({
             patterns: [
-                './node_modules/@microsoft/vscode-azext-azureutils/resources/**/*.svg',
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -71,7 +71,7 @@ const config = {
         // Copy some needed resource files from external sources
         new CopyPlugin({
             patterns: [
-                './node_modules/@microsoft/vscode-azext-azureutils/resources/**/*.svg',
+                './node_modules/@microsoft/vscode-azext-azureutils/resources/*.svg',
```
bwateratmsft, line 74: This is unrelated, but recently `@microsoft/vscode-azext-azureutils` started bundling a lot more Azure icons. We only need the ones at the root, not for every service.

karolz-ms (approved):

PR #253, opened 2025-10-06
Feature/tree view search

FilipCondac:
## Description

This PR adds search/filter functionality to all tree view sections in the Container Tools extension, making it easier for users to find specific containers, images, networks, volumes, contexts, and registries when working with large numbers of resources.

## Motivation

I work in an env with quite a few containers and it was always quite frustrating trying to find the right one quickly.

To clear the filter:

1. Click the search icon again
2. Either:
   - Press Enter with empty input, or
   - Select "Clear Filter" option

### Search Across Multiple Properties

The filter searches through:

- Container/Image/Volume names
- Tags
- IDs
- Status
- Description fields
- And more!

Example: Searching "running" will show all running containers.

### Case-Insensitive Search

- "NGINX" = "nginx" = "Nginx"
- Type in lowercase for faster searching

### Partial Matching

- "ngi" matches "nginx"
- "post" matches "postgres"
- "8080" matches containers with port 8080

### Multiple Views, Independent Filters

Each view maintains its own filter:

- Filter containers by "web"
- Filter images by "node"
- Both filters remain active simultaneously

### Example 1: Find Development Containers

```
1. Click search in Containers view
2. Type: "dev"
3. See: dev_frontend, dev_backend, dev_database
```

### Example 2: Find Images by Version

```
1. Click search in Images view
2. Type: "14"
3. See: postgres:14, node:14-alpine, python:3.14
```

### Example 3: Find Networks by Name

```
1. Click search in Networks view
2. Type: "bridge"
3. See: bridge, custom_bridge, my_bridge_network
```

### Example 4: Find Stopped Containers

```
1. Click search in Containers view
2. Type: "exited"
3. See: All containers with "exited" status
```

### Want to See Everything Again?

- Click search icon
- Press Enter with empty input
- Description will clear and all items return

### Files Changed

**New Files:**
- `src/commands/filterTree.ts` - Core filter logic and commands

**Modified Files:**
- `src/tree/LocalRootTreeItemBase.ts` - Integrated filtering into tree item loading
- `src/commands/registerCommands.ts` - Registered 6 new filter commands
- `package.json` - Added command definitions and menu contributions
- `package.nls.json` - Added localization strings for filter commands

## How to Use

1. Click the 🔍 search icon in any tree view header
2. Type search term in the Quick Pick input
3. Press Enter to apply the filter
4. View updates to show only matching items
5. Description shows: "🔍 Filtered: 'search-term'"
6. Click search again and press Enter (empty) to clear

### Filter Algorithm
- Searches across multiple properties: label, description, and all configured properties
- Case-insensitive matching for better UX
- Partial string matching for flexibility
- Filter state stored per tree view prefix

### Telemetry
- Tracks filter application/clearing actions
- Records filter text length
- Counts filtered item results

### Design Decisions
- Positioned search button between runtime selector and settings gear
- Filter persists during session but resets on reload (expected behavior)
- Multi-property search provides comprehensive results

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Comments added for complex logic
- [x] All files properly formatted
- [x] No console errors or warnings
- [x] Telemetry tracking added
- [x] Localization strings added
- [x] Manual testing completed across all views
- [x] Works on macOS/Windows/Linux (tested on: [specify])

## Additional Notes

This implementation uses VS Code's native Quick Pick API to provide a search experience similar to the Command Palette. While it differs from a persistent search bar (like in the Marketplace), it offers several advantages:
- Familiar to VS Code users
- Keyboard accessible
- No custom UI complexity
- Consistent with extension command patterns

Future enhancements could include:
- Search history/suggestions
- Regular expression support
- Property-specific filters (e.g., "status:running")
- Save/load filter presets

## Demo

---

**Thank you for reviewing this contribution! I'm happy to address any feedback or make adjustments as needed.** 🚀


isidorn:
@benibenj should be able to help with this

FilipCondac:
> @bwateratmsft they keybiding is `Alt + Ctrl + F`. You can change the keybinding for `list.find`

This wasn't defined by default in my VSCode settings. Had to create a custom keybind which does now allow me to activate the search bar feature.

```
[
  {
    "key": "cmd+f",
    "command": "list.find",
    "when": "listFocus && !inputFocus"
  }
]

```

FilipCondac:
I get the point about `list.find`, but this feature goes a bit beyond that. `list.find` only highlights matches, it doesn’t actually filter the list or reduce clutter. Plus, it’s not enabled by default and requires a keybind (not sure if all users?), so some users won't know that exists.

bwateratmsft:
> I get the point about `list.find`, but this feature goes a bit beyond that. `list.find` only highlights matches, it doesn’t actually filter the list or reduce clutter. Plus, it’s not enabled by default and requires a keybind (not sure if all users?), so some users won't know that exists.

I agree, this PR definitely goes beyond the features that `list.find` provides. There is a little button that toggles between highlighting and filtering in `list.find` though. However, I did not have to set the keybinding for it--are you on Mac or Windows?

FilipCondac:
> I have a few general pieces of feedback.
> 
> * I'd prefer for the search icon to change to `$(search-stop)` and a single click to that to clear the search terms. This could be implemented similar to how we do the commands `vscode-containers.images.showDangling` and `vscode-containers.images.hideDangling`. Unfortunately, this doubles the number of commands which is a bummer.
> * Fuzzy search is pretty deeply ingrained into the habits of VSCode users, I'm wondering if there's a reasonably simple way we can support that. The builtin search/filter feature does include that.
> * We could probably get away with supporting this only for images and containers. It's not common for other types (networks, volumes etc.) to have so many instances.
> 
> In the past, we have considered supporting a group-by-arbitrary-label feature. Right now for example, the default container grouping is by compose project, which comes from labels on the containers. #104 is related but not exactly the same.
> 
> Would your use case be served as well by being able to set arbitrary labels on the containers, and use them for grouping? While not the same as filtering, that would at least allow you to group in any way you want.

I think the `group-by-arbitrary-label` feature would be also great addition, especially for developers working in monorepos or microservice heavy environments. It would be genuinely useful to visually group related services.

However, I personally see filtering as the more universally valuable feature. Speaking for myself (and likely many other developers), I often prioritise features that just work out of the box. Setting up labels adds a layer of manual organisation. In practice it could be time consuming for big projects but beneficial for those who love organisation.

bwateratmsft:
That makes sense! By the way, your formatter went a little mad with power :smile:

Can you revert the formatted lines? Don't want to put you on the `git blame` for everything

FilipCondac:
- Removed 

> That makes sense! By the way, your formatter went a little mad with power 😄
> 
> Can you revert the formatted lines? Don't want to put you on the `git blame` for everything

Oh phew, slightly embarrassing but all fixed. One wrong command is all it takes sometimes.

bwateratmsft:
What's the error and in what phase of building? If it's happening during `npm run test`, it's probably due to the current Azure outage.

FilipCondac:
@bwateratmsft Forgot to ask if you want a demo video of the new iteration.

FilipCondac:
Demo for fuzzy matches and new one click clear

https://github.com/user-attachments/assets/b796812a-7438-486c-97d3-3e17708f285c

src/commands/filterTree.ts
bwateratmsft: Can you update the spacing in this file to 4 spaces per indent?

src/commands/filterTree.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  const quickPick = vscode.window.createQuickPick();
+  quickPick.placeholder = `Filter ${treePrefix}... (Press Enter to apply, Esc to cancel)`;
+  quickPick.value = currentFilter.filterText;
+  quickPick.title = `Filter ${capitalize(treePrefix)}`;
+
+  if (currentFilter.isActive) {
+    quickPick.items = [
+      {
+        label: "$(clear-all) Clear Filter",
```
bwateratmsft: This label gets used twice (here plus comparison below), I'd suggest moving either the string itself or the entire quick pick item into a constant. Also, `vscode.l10n.t()` applies here too.

src/commands/filterTree.ts
bwateratmsft: Can you also merge the latest main, and do `npm run lint` and fix any issues that arise? Since you originally authored, we have adopted some stricter linting rules that might be triggered.

PR #266, opened 2025-10-16
Bump Dockerfile language service plus a few minor changes

bwateratmsft:


src/extension.ts
```diff
@@ -340,7 +331,7 @@ function activateComposeLanguageClient(ctx: vscode.ExtensionContext): void {
             )
         );
 
-        const debugOptions = { execArgv: ["--nolazy", "--inspect=6009"] };
+        const debugOptions = { execArgv: ["--nolazy", "--inspect=6010"] };
```
bwateratmsft: Our compose lang service was trying to listen on the same port as the Dockerfile one, when debugging the extension. Not useful.

patverb (approved):

PR #344, opened 2026-01-20
Remove Copilot tools in favor of config tool only

### LINKED ISSUE #286: ~~More~~ LESS Copilot tools!
Let's add even ~~MORE~~ LESS copilot tools!

- [x] Configuration readout? i.e. a tool that can tell Copilot what the current container client config is, e.g. Docker vs Podman, env vars, etc.
- ~~Compose up/down/stop/etc.~~
- ~~Attach shell / maybe attach VSCode~~
- ~~Docker build~~

bwateratmsft:
This PR removes the Copilot tools previously added, in favor of one simple LM tool that provides configuration information. This allows Copilot to generate its own CLI commands, taking full advantage of the CLI's power, without losing the critical configuration information.

This also reduces the tool count.

Fixes #286

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,8 +5,7 @@
         "target": "es2022",
         "outDir": "out",
         "lib": [
-            "es2022",
-            "dom"
+            "es2022"
         ],
         "sourceMap": true,
         "rootDir": "src",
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -5,8 +5,7 @@
         "target": "es2022",
         "outDir": "out",
         "lib": [
-            "es2022",
-            "dom"
```
bwateratmsft: Don't need this anymore

patverb (approved):

PR #372, opened 2026-02-02
Fix telemetry bundling issue

### LINKED ISSUE #367: Extension telemetry is not being sent by 2.4.0
Boring analysis: the reason it was happening was because we're using `keepNames` in esbuild, to make it not mangle class/function names (which we need if we want useful call stacks). Esbuild writes code into the bundle that, at load time (which is run time), tries to set the `name` property on the minified classes/functions to the original name.
 
Some of the AppInsights packages are freezing some of their stuff, which makes it impossible to set the `name` property. The error that was thrown (and then caught and eaten) would resemble:

```
TypeError: Cannot redefine property: name
at defineProperty (<anonymous>)
at __name (d:\\vscode-containers\\dist\\extension.bundle.js:8:33)
at node_modules/@microsoft/1ds-core-js/dist-es5/AppInsightsCore.js (d:\\vscode-containers\\dist\\extension.bundle.js:16157:5)
at __init (d:\\vscode-containers\\dist\\extension.bundle.js:10:56)
at node_modules/@microsoft/1ds-core-js/dist-es5/Index.js (d:\\vscode-containers\\dist\\extension.bundle.js:16819:5)
at __init (d:\\vscode-containers\\dist\\extension.bundle.js:10:56)
at d:\\vscode-containers\\dist\\extension.bundle.js:20303:57
at async getAICore (d:\\vscode-containers\\dist\\extension.bundle.js:20303:21)
at async oneDataSystemClientFactory (d:\\vscode-containers\\dist\\extension.bundle.js:20343:29)
```
 
The end result was that loading the AppInsights packages failed silently, and the AppInsights reporter never got built--so when we logged telemetry events, they were just screamed into the void.

I considered three options. The first was to build the extension as ESM (#366), but unfortunately this just didn't work at all.

The second was to turn off `keepNames`...but that would negatively impact both telemetry and the issue reporter by giving us mangled names that we would have to translate after-the-fact if we wanted to investigate.

The third was Opus' idea (nice job AI!). Basically, we build the `@vscode/extension-telemetry` package into its own bundle, with `keepNames` off, and then build everything else (extension and other dependencies) into the main bundle, with an alias to rewrite imports to `@vscode/extension-telemetry` to point to the other bundle.

bwateratmsft:
Fixes #367

Verified the fix by checking Kusto--events are being sent from me

### DIFF
--- a/esbuild.mjs
+++ b/esbuild.mjs
@@ -5,13 +5,13 @@
 
 import { autoEsbuildOrWatch, autoSelectEsbuildConfig } from '@microsoft/vscode-azext-eng/esbuild';
 
-const baseConfig = autoSelectEsbuildConfig();
+const { extensionConfig, telemetryConfig } = autoSelectEsbuildConfig();
 
 /** @type {import('esbuild').BuildOptions} */
 const finalConfig = {
-    ...baseConfig,
+    ...extensionConfig,
     entryPoints: [
-        ...baseConfig.entryPoints,
+        ...extensionConfig.entryPoints,
         {
             in: './node_modules/dockerfile-language-server-nodejs/lib/server.js',
             out: 'dockerfile-language-server-nodejs/lib/server',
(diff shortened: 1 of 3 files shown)


patverb (approved):

xmama1230 (commented):
k

PR #381, opened 2026-02-06
Take advantage of new changes to compose language service package

bwateratmsft:


src/utils/lazyPackages.ts
```diff
@@ -26,6 +26,15 @@ export async function getHandlebars() {
     return await handlebarsLazy.value;
 }
 
+const languageClientLazy = new Lazy(async () => {
+    const module = await import('vscode-languageclient/node');
+    // The module is CJS; when bundled to ESM by esbuild the exports end up on `default`. Fall back to the module itself for unbundled usage.
+    return (module as unknown as typeof module & { default?: typeof module }).default ?? module;
```
bwateratmsft: I don't feel good about this but it works and I'm not sure of an alternative.

PR #451, opened 2026-04-21
Create codeql.yml

2congamaidau-create:


bwateratmsft:
CodeQL is already set up for this repo through some internal systems, so we don't need this--but thank you for the contribution!

PR #473, opened 2026-05-12
Do AppSvc deployments through AppSvc extension

### LINKED ISSUE #117: Move image deployment to App Service extension
Similar to how we want to do ACA implementation, let's move the App Service deployment out into the AppSvc extension. A soft dependency in the Docker extension can call a command in AppSvc.

Related to microsoft/vscode-docker#3535

### LINKED ISSUE #466: Need to use new AppSvc audience for Kudu calls
See https://github.com/microsoft/vscode-azuretools/pull/2258

We will need to adopt the latest AppSvc package, and inventory anything else that's relevant (Log streaming? Deploy to ACA as well?)

bwateratmsft:
Corresponds to https://github.com/microsoft/vscode-azureappservice/pull/2871

Closes #117 and indirectly closes #466


src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -3,69 +3,158 @@
...
+            tag,
+            username: logInInfo.username,
+            secret: logInInfo.secret,
+        };
+    }
+
+    // Don't wait
+    void vscode.commands.executeCommand('appService.deployImageApi', commandOptions);
+}
+
```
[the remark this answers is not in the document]
bwateratmsft: We are already validating the version so I'm not worried.

src/commands/registries/azure/deployImageToAzure.ts
```diff
@@ -3,69 +3,113 @@
...
+    repositoryName: string;
+    tag: string;
+    username?: string;
+    secret?: string;
+    acrRegistry?: AcrRegistryPropertiesContract;
 }
 
 export async function deployImageToAzure(context: IActionContext, node?: UnifiedRegistryItem<CommonTag>): Promise<void> {
+    // Assert installation of the App Service extension
+    if (!isExtensionInstalledAndVersionCompatible(appServiceExtensionId, minimumAppServiceExtensionVersion)) {
```
patverb, line 41: I assume vs code people wont really care that they now need to install an extension? Given they probably already have it installed if you are deploying to app services anyways.

src/utils/installExtension.ts
```diff
@@ -1,33 +1,63 @@
...
+        );
+
+    const action: vscode.MessageItem = {
+        title: isUpdate ? vscode.l10n.t('Update') : vscode.l10n.t('Install'),
+    };
+
+    const result = await context.ui.showWarningMessage(message, { modal: true }, action, DialogResponses.cancel);
+
+    if (result === action) {
+        await vscode.commands.executeCommand('extension.open', extensionId);
```
patverb, line 56: Do we need to open the marketplace page if we are going to install? Is there any scenario where open could fail but install would still work? Probably not.

src/utils/installExtension.ts
```diff
@@ -1,33 +1,63 @@
...
+            minimumVersion
+        )
+        : vscode.l10n.t(
+            'Version {0} or higher of the {1} extension is required to deploy to {1}. Would you like to install it now?',
+            minimumVersion,
+            extensionDisplayName
+        );
+
+    const action: vscode.MessageItem = {
+        title: isUpdate ? vscode.l10n.t('Update') : vscode.l10n.t('Install'),
```
patverb, line 50: Should we check that the min version is even possible? IF its not we could just ask them to update, open the store and then next install they keep getting the same "update" dialog. Now that would be a failure on our part but if we have to remove a new version or something it could happen and might be nice to have an exit for the error path.

src/utils/installExtension.ts
```diff
@@ -1,33 +1,63 @@
...
+import * as vscode from 'vscode';
+
+export function isExtensionInstalledAndVersionCompatible(extensionId: string, minimumVersion: string): boolean {
+    const extension = vscode.extensions.getExtension(extensionId);
+
+    if (!extension?.packageJSON?.version) {
+        return false;
+    }
+
+    const extensionVersion = semver.parse(extension.packageJSON.version) ?? semver.coerce(extension.packageJSON.version);
```
patverb, line 17: Does semver handle pre prelease versioning comparison wells? like 2.23.0-alpha?

src/utils/installExtension.ts
```diff
@@ -1,33 +1,63 @@
...
+
+    const action: vscode.MessageItem = {
+        title: isUpdate ? vscode.l10n.t('Update') : vscode.l10n.t('Install'),
+    };
+
+    const result = await context.ui.showWarningMessage(message, { modal: true }, action, DialogResponses.cancel);
+
+    if (result === action) {
+        await vscode.commands.executeCommand('extension.open', extensionId);
+        await vscode.commands.executeCommand('workbench.extensions.installExtension', extensionId);
```
patverb, line 57: Should we await the extension being activated?

src/utils/installExtension.ts
```diff
@@ -1,33 +1,63 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { DialogResponses, IActionContext, UserCancelledError } from '@microsoft/vscode-azext-utils';
+import * as semver from 'semver';
+import * as vscode from 'vscode';
+
+export function isExtensionInstalledAndVersionCompatible(extensionId: string, minimumVersion: string): boolean {
+    const extension = vscode.extensions.getExtension(extensionId);
```
patverb, line 11: Will this handle disabled extensions? What happens when a disabled extension is asked to update? Should we prompt to enable?

src/commands/registries/azure/deployImageToAzure.ts
bwateratmsft: Yeah, it's a fairly common pattern for us to have soft dependencies like this. User are accustomed to it. This is already how deploy to ACA works.

src/utils/installExtension.ts
bwateratmsft: Yes and no. Strictly speaking we do not need to open the marketplace page at all, but I feel like it's a good user-friendly practice to do it anyway.

src/utils/installExtension.ts
bwateratmsft: I'm not sure of any way to check that. Definitely nothing built-in and I feel like removing a new version is unlikely and, even if it does happen, short-term.

src/utils/installExtension.ts
bwateratmsft: Opus says yes but I'll test it :smile:

src/utils/installExtension.ts
bwateratmsft: We don't actually need to, since VSCode will internally await it when we call the command to start the deployment.

src/utils/installExtension.ts
bwateratmsft: AFAIK there isn't a way to tell if an extension is not installed vs. installed but disabled. I'm not exactly sure what happens if we try this with the extension disabled--I will test and reply.

patverb (approved):

src/utils/installExtension.ts
bwateratmsft: Looks like both `parse` and `coerce` handle it well. `parse` preserves pre-release info, `coerce` excludes it. In either case, we will correctly see that e.g. `'2.4.9-alpha' >= '2.4.9'`

src/utils/installExtension.ts
bwateratmsft: Ok, as I suspected, it just appears as being uninstalled to us. Opening the extension page does help clue users in though, so I think this is good enough for now.

PR #492, opened 2026-05-28
Visualize running container counts on tree root and group nodes

### LINKED ISSUE #485: [FEAT] Show if some or all containers are running
With many containers it can be hard to see if and where containers are running. On the collapsed tree visualise the state of some or all containers:

## Example

Image

In this example all containers in wp-test are running but at first glance that is not visual;

Image

The icon collapsed or not show show that some or all containers are actively running

SATYAM-PRATIBHAN:
## Description

This PR improves container visibility in the tree view when nodes are collapsed or grouped, which is especially helpful in environments with many containers.

Dynamic status summaries were added to the description field of the **Containers** root node and individual container group nodes (such as Docker Compose projects). This makes it possible to see container state at a glance without manually expanding nodes.

## Key Changes

- **Root Node:** Displays `(X running)` next to the **Containers** label when one or more containers are active.
- **Group Nodes:** Displays `(X/Y running)` next to group names, showing the number of running containers versus the total number of containers in that group.
- **Base Class Refactor:** Updated `LocalRootTreeItemBase` and `LocalGroupTreeItemBase` to safely expose item data to subclasses, enabling these summaries while following existing architectural patterns.
- **Unit Tests:** Added a new test suite at `src/test/tree/containers/ContainerTreeVisualization.test.ts` to verify the description logic for both root and group nodes.

## Related Issue

Fixes #485 

## Visuals

Before:
Screenshot 2026-05-29 at 00 33 02

After:
Screenshot 2026-05-29 at 00 28 40

## Testing

- [x] Manual verification: Verified in Extension Development Host that descriptions update correctly when containers start and stop.
- [x] Type checking: `npm run build:check` passed successfully.


SATYAM-PRATIBHAN:

src/tree/LocalGroupTreeItemBase.ts
```diff
@@ -10,7 +10,7 @@ import { CommonProperty } from "./settings/CommonProperties";
 export abstract class LocalGroupTreeItemBase<TItem extends AnyContainerObject, TProperty extends string | CommonProperty> extends AzExtParentTreeItem {
     public declare readonly parent: LocalRootTreeItemBase<TItem, TProperty>;
     public readonly group: string;
-    private _items: TItem[];
+    protected _items: TItem[];
```
bwateratmsft: Here and the other class; I would add a protected readonly accessor instead of making the private field protected. This would also resolve the lint warning. Something like this:

```typescript
    protected get items(): readonly TItem[] {
        return this._items;
    }
```

SATYAM-PRATIBHAN: makes sense, thanks for the help :)

bwateratmsft:
@fiveisprime / @SATYAM-PRATIBHAN what do you think about only showing this additional description if there are non-running containers? i.e. less than 100% running

extensions/vscode-containers/src/tree/LocalRootTreeItemBase.ts
```diff
@@ -69,6 +69,10 @@ export abstract class LocalRootTreeItemBase<TItem extends AnyContainerObject, TP
     private _cachedItems: TItem[] | undefined;
     private _currentDockerStatus: DockerStatus;
 
+    protected get currentItems(): TItem[] | undefined {
```
bwateratmsft, line 72: To protect against mistakes, I'd do:
```suggestion
    protected get currentItems(): Readonly<TItem[]> | undefined {
```

Or maybe there's `ReadonlyArray<TItem>`, I can't remember

extensions/vscode-containers/src/tree/LocalGroupTreeItemBase.ts
```diff
@@ -13,6 +13,10 @@ export abstract class LocalGroupTreeItemBase<TItem extends AnyContainerObject, T
     private _items: TItem[];
     private _childTreeItems: AzExtTreeItem[];
 
+    protected get items(): TItem[] {
```
bwateratmsft, line 16: To protect against mistakes, I'd do:

```suggestion
    protected get items(): Readonly<TItem[]> {
```

Or maybe there's `ReadonlyArray<TItem>`, I can't remember

bwateratmsft (changes requested):
Can you also include a screenshot of the label on the Containers root?

bwateratmsft:
Closing for now due to inactivity, but we can reopen later if needed.

PR #518, opened 2026-07-04
Dangling images are showing up again

varmiral:
Since Docker v29.0.0, dangling images are not included by default, so the `all` option was added
The `all` option should also display intermediate images, but the default BuildKit does not create them


bwateratmsft (approved):

patverb (approved):

bwateratmsft:
This is now released in v2.5.0 of the Container Tools extension.

Issue #501, opened 2026-06-11
Support new WSL "native" container runtime


Clockwork-Muse:
Microsoft announced a [WSL "native" container runtime](https://learn.microsoft.com/en-us/windows/wsl/wsl-container) (not requiring Docker or Podman) as Build on 2026-06-03.
It would be nice to support this runtime.

MrHinsh:
This is what I came here to say. Im not using docker and only using WSL Containers...

Clockwork-Muse:
> This is what I came here to say. Im not using docker and only using WSL Containers...

WSL Containers aren't in (official) public preview yet, so unless you've built and installed them from source I'm curious what it is you're currently using...?

Diegorro98:
Is already on public preview!

craigloewen-msft:
This is already available today, you can set `wslc` as your preferred container tool on  0.462.0 and higher:
https://devblogs.microsoft.com/commandline/wsl-container-is-now-available-for-public-preview/#vs-code-dev-containers

<img width="477" height="98" alt="Image" src="https://github.com/user-attachments/assets/fc429a83-d6ae-4902-a409-45c80bb6d21d" />

bwateratmsft:
> This is already available today, you can set `wslc` as your preferred container tool on 0.462.0 and higher: https://devblogs.microsoft.com/commandline/wsl-container-is-now-available-for-public-preview/#vs-code-dev-containers
> 
[...]

PR #526, opened 2026-07-16
Add wslc support

patverb:
###  Add  wslc  (WSL Container CLI) as a Windows-only container runtime (preview)

 Description

 Registers the new  WslcClient  from  @microsoft/vscode-container-client  as a selectable, Windows-only container runtime, exposed as "WSLC (Windows  only, preview)". It's treated as an alternative to Docker/Podman, mirroring the existing auto-configurable client pattern.

 Depends on microsoft/vscode-docker-extensibility# ( @microsoft/vscode-container-client@^0.6.0 ).

###  What's added / changed

 •  AutoConfigurableWslcClient  — mirrors the Docker/Podman auto-configurable clients, but pins the command to  wslc  (ignores the shared     containers.containerCommand  override, since wslc is a specific named binary).
 •  extension.ts  — registers the wslc client only on Windows ( isWindows()  gate), and wires up  reconfigure .
 •  RuntimeManager  — if  wslc  is selected on a non-Windows host, rejects with a clear "WSLC is only available on Windows" error instead of a confusing    timeout.

###  .NET SDK container task support

 The .NET SDK's  LocalRegistry  publish target only supports Docker/Podman. To support wslc, the SDK run task now:

 1. Publishes the image to a temporary tar archive ( -p:ContainerArchiveOutputPath=<tmp>\<image>-<guid>.tar ) instead of  -p:LocalRegistry ,
 2. Loads it with  wslc load --input <tar> ,
 3. Runs it — then deletes the temp tar in a  finally  (GUID-named to avoid collisions).

 Docker/Podman continue to use  LocalRegistry  unchanged.

#501

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.5.0 - Unreleased
+### Added
+* Added support for `wslc` (Windows Subsystem for Linux Container CLI) as an alternative container runtime. Windows-only.
+
 ## 2.4.5 - 28 May 2026
 ### Changed
 * Deployments to Azure App Service are now performed through the [Azure App Service extension](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azureappservice), simplifying the deployment flow and reducing duplicated code. [#473](https://github.com/microsoft/vscode-containers/pull/473)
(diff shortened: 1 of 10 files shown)


bwateratmsft:
Closing in favor of #561

PR #529, opened 2026-07-16
Simplify runtime management and drop Compose V1 support

### LINKED ISSUE #339: Drop support for Compose V1
Compose V1 has been out-of-support for about 2.5 years now. I think it's OK at this point to drop support for it as well. Essentially, this means we will only use `docker compose` / `podman compose` going forward as our orchestrator command.

bwateratmsft:
## Summary

Simplifies container/orchestrator runtime setup and drops support for Compose V1 (`docker-compose`) in favor of V2 (`docker compose`).

### Runtime simplification
- Deletes the five `AutoConfigurable*Client` wrappers. The base `DockerClient` / `PodmanClient` / `DockerComposeClient` / `PodmanComposeClient` are now registered directly.
- Moves reconfiguration into the `RuntimeManager` base class, which now owns the `onDidChangeConfiguration` subscription and a `reconfigureClient` hook, replacing the per-client `reconfigure()` and the config event previously wired up in `registerContainerClients`.
- `RuntimeManager` constructor now takes `(defaultClientId, clientSettingName, overrideCommandSettingName)` and resolves the default client by ID from the registry instead of a `getDefaultClient()` override.
- Moves `isComposeV2ableOrchestratorClient` into `OrchestratorRuntimeManager` (its old home was deleted); consumers in `selectCommandTemplate.ts` and `containersConfigTool.ts` updated to import from the new location.

### Drop Compose V1
- Removes the slow `docker` vs `docker-compose` autodetection. Orchestrator clients now always run in V2 mode (`composeV2 = true`).
- `OrchestratorRuntimeManager.reconfigureClient` normalizes any V1-style (`docker-compose`) or explicit V2 (`docker compose`) override down to the base command (`docker`), since composeV2 appends `compose` itself.

### `defaultCommandName` on the container client
- Adds an immutable `defaultCommandName` to the `ClientIdentity` contract in `packages/vscode-container-client`, implemented in `ConfigurableClient` from the constructor''s command name.
- This lets `RuntimeManager.reconfigureClient` reset a client back to its original default: `client.commandName = this.getOverrideSettingValue() || client.defaultCommandName;`
- Because the client is now vendored as a workspace package, the change lands here directly — no publish/version bump needed, and the earlier temporary runtime shim has been removed.

## Validation
- Package (`vscode-container-client`): build clean, lint clean, 120 tests passing
- Extension (`vscode-containers`): `tsc --noEmit` clean, `eslint src` clean, esbuild clean, 148 tests passing


extensions/vscode-containers/src/runtimes/RuntimeManager.ts
```diff
@@ -61,9 +74,21 @@ export abstract class RuntimeManager<TClient extends ClientIdentity> extends vsc
...
 
-    protected abstract getDefaultClient(): TClient;
+    protected reconfigureClient(client: TClient): void {
+        // TEMPORARY: `defaultCommandName` is being added to `ClientIdentity` in @microsoft/vscode-container-client.
+        // Until that change is published and this extension's dependency is bumped, cast to reach it and fall back
+        // to the client's current command name (its constructor default) since the property isn't present at runtime yet.
+        // Once the dependency is updated, replace the next two lines with:
+        //   client.commandName = this.getOverrideSettingValue() || client.defaultCommandName;
+        const defaultCommandName = (client as unknown as { defaultCommandName?: string }).defaultCommandName ?? client.commandName;
+        client.commandName = this.getOverrideSettingValue() || defaultCommandName;
```
bwateratmsft: Once `defaultCommandName` is added to `ClientIdentity` in `@microsoft/vscode-container-client` and the dependency here is bumped, apply this suggestion to drop the temporary cast + runtime fallback and restore the intended one-liner:

```suggestion
        client.commandName = this.getOverrideSettingValue() || client.defaultCommandName;
```

extensions/vscode-containers/src/runtimes/OrchestratorRuntimeManager.ts
```diff
@@ -5,31 +5,32 @@
...
-        }
+        if (isComposeV2ableOrchestratorClient(client)) {
+            // Compose V1 (`docker-compose`) is no longer supported, so these clients always run in V2 mode,
+            // where `compose` is appended as the first argument to the base command (e.g. `docker compose`).
+            client.composeV2 = true;
 
-        return orchestratorClient;
+            // Normalize a V1-style (`docker-compose`) or explicit V2 (`docker compose`) override down to the
+            // base command, since composeV2 appends `compose` itself and we'd otherwise get `docker-compose compose`.
+            client.commandName = client.commandName.replace(/[-\s]+compose\s*$/i, '');
```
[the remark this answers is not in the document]
bwateratmsft: Accepted. The normalization regex now strips an optional .exe suffix, so docker-compose.exe and podman-compose.exe normalize correctly to the base command under composeV2.

extensions/vscode-containers/src/runtimes/RuntimeManager.ts
```diff
@@ -3,17 +3,33 @@
...
+        this.onDidChangeConfigurationDisposable = vscode.workspace.onDidChangeConfiguration(async e => {
+            if (e.affectsConfiguration(`${configPrefix}.${this.overrideCommandSettingName}`)) {
+                return callWithTelemetryAndErrorHandling('vscode-containers.command.changed', (context: IActionContext) => {
+                    for (const client of this._runtimeClients.values()) {
+                        this.reconfigureClient(client);
+                    }
+                });
+            }
+        });
+    }
```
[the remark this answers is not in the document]
bwateratmsft: Partially accepted: the listener is now non-async, calls void callWithTelemetryAndErrorHandling(...), and no longer takes an unused context parameter. I intentionally kept telemetry and error display active because the owner wants to track command-change events, so I did not add the suppress flags.

extensions/vscode-containers/src/runtimes/RuntimeManager.ts
```diff
@@ -34,18 +51,14 @@ export abstract class RuntimeManager<TClient extends ClientIdentity> extends vsc
...
 
-    public get runtimeClients(): Array<TClient> {
-        return Array.from(this._runtimeClients.values());
-    }
-
     public async getClient(): Promise<TClient> {
         const config = vscode.workspace.getConfiguration(configPrefix);
         const runtimeClientId = config.get<string | undefined>(this.clientSettingName);
 
         let runtimeClient: TClient;
```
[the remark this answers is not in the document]
bwateratmsft: Accepted: getClient() now resolves through an effective client id with waitForClientToBeRegistered. This fixes both default-client ordering and the misleading error message.

extensions/vscode-containers/src/runtimes/RuntimeManager.ts
```diff
@@ -61,9 +68,18 @@ export abstract class RuntimeManager<TClient extends ClientIdentity> extends vsc
...
 
-    protected abstract getDefaultClient(): TClient;
+    protected reconfigureClient(client: TClient): void {
+        // Fall back to the client's current command name so third-party clients registered via the public
+        // extension API (which may predate `defaultCommandName` on the ClientIdentity contract) are never
+        // wiped to `undefined`.
+        client.commandName = this.getOverrideSettingValue() || client.defaultCommandName || client.commandName;
+    }
+
+    private getOverrideSettingValue(): string | undefined {
```
patverb, line 78: This seems like it could be problematic if a user has a custom path for one client but not another. If they need to set it for say docker.  Then that custom command will be used for every runtime including podman/wslc. Is there a way we can make the override runtime specific? Its not a new bug but something that could cause issues the more runtimes we add.

extensions/vscode-containers/src/runtimes/OrchestratorRuntimeManager.ts
```diff
@@ -5,31 +5,32 @@
...
-        }
+        if (isComposeV2ableOrchestratorClient(client)) {
+            // Compose V1 (`docker-compose`) is no longer supported, so these clients always run in V2 mode,
+            // where `compose` is appended as the first argument to the base command (e.g. `docker compose`).
+            client.composeV2 = true;
 
-        return orchestratorClient;
+            // Normalize a V1-style (`docker-compose`) or explicit V2 (`docker compose`) override down to the
+            // base command, since composeV2 appends `compose` itself and we'd otherwise get `docker-compose compose`.
+            client.commandName = client.commandName.replace(/[-\s]+compose(\.exe)?\s*$/i, '');
```
patverb: Is it possible for the command name to be a full path?  If someone has a custom path like `C:\tools\docker-compose.exe` Then we change it to `C:\tools\docker` and that wont work if its windows or if the non compose binary not next to it.

extensions/vscode-containers/src/runtimes/OrchestratorRuntimeManager.ts
```diff
@@ -5,31 +5,32 @@
...
-    public readonly onOrchestratorRuntimeClientRegistered = this.runtimeClientRegisteredEmitter.event;
-
     public constructor() {
-        super('orchestratorClient');
+        super(DockerComposeClient.ClientId, 'orchestratorClient', 'composeCommand');
     }
 
-    public override async getClient(): Promise<IContainerOrchestratorClient> {
-        const orchestratorClient = await super.getClient();
+    protected override reconfigureClient(client: IContainerOrchestratorClient): void {
```
patverb, line 20: Do we want to be attempting to automatically fix this or should we just bubble up an error and tell the user to fix it themselves?

patverb (dismissed):

extensions/vscode-containers/src/runtimes/OrchestratorRuntimeManager.ts
bwateratmsft: That's a good question. In the past we've done the fix but we don't necessarily have to...

extensions/vscode-containers/src/runtimes/OrchestratorRuntimeManager.ts
bwateratmsft: It is, but I'm not sure if PATH resolution would understand that `C:\tools\docker` means `C:\tools\docker.exe`. Worth a test.

extensions/vscode-containers/src/runtimes/RuntimeManager.ts
bwateratmsft: It was this way in the past, at least somewhat on purpose. The thinking is that the `commandName` setting wins over *everything*, regardless of what client you choose. For example, some users have Podman installed and configured, and `docker` is aliased to `podman`. I can imagine some of them would be accustomed to running `docker` commands, and would want the extension to do the same--even though the client is really Podman and we should use Podman-shaped arguments.

extensions/vscode-containers/src/runtimes/OrchestratorRuntimeManager.ts
bwateratmsft: (Becomes moot if we do [comment 3 below](https://github.com/microsoft/vscode-containers/pull/529#discussion_r3615995958))

extensions/vscode-containers/src/runtimes/RuntimeManager.ts
patverb: Thats a good point. I was thinking this setting was more for "I have it installed in a specific location and ill tell you how to run it". But if its more intended for "run this command, dont run what vscode wants" that makes more sense. 

The only concern would be like you mentioned is the arguments shape might be different for docker/podman/wslc and the returned results wont parse.

But also these settings are right next to each other and it looks obvious what it does so its on the user if they get it wrong

extensions/vscode-containers/src/runtimes/OrchestratorRuntimeManager.ts
patverb: Yeah path wouldn't know how to call that. although windows specifically has a "docker" file that redirects to the exe. We could always just check if the path is rooted. If it is we dont try to redirect this and give an error.

extensions/vscode-containers/src/runtimes/OrchestratorRuntimeManager.ts
patverb: Im impartial. Depending on our usage of v1. If we have done this is the past and no one cared then its fine. Otherwise im curious if some people would complain? I doubt it given I bet everyone has a docker new enough with compose and it doesnt really matter. The only concern is bugs that are our fault vs telling the user its broke you go fix it yourself.

bwateratmsft: Compose v1 is super deprecated at this point...like 3 years dead. I think I could live with breaking it...

extensions/vscode-containers/src/runtimes/RuntimeManager.ts
bwateratmsft: Yeah, this is definitely an at-your-own-risk setting.

extensions/vscode-containers/src/runtimes/OrchestratorRuntimeManager.ts
bwateratmsft: Good catch. We no longer rewrite the command at all, so a rooted path like "C:\tools\docker-compose.exe" is never silently turned into "C:\tools\docker". Instead reconfigureClient detects the trailing compose subcommand and throws an actionable error asking for the base command only.

bwateratmsft: Confirmed: PATH would not resolve a rewritten "C:\tools\docker", which is exactly why we stopped rewriting. reconfigureClient now throws an actionable error pointing at the setting instead of guessing.

bwateratmsft: Yep, went with the error approach below, so this is now moot. No rewriting happens; a command that still ends in the compose subcommand throws a clear message.

bwateratmsft: Took a simpler version of this: rather than special-casing rooted paths, any command still ending in the compose subcommand (V1 "docker-compose" or explicit "docker compose") throws an actionable error. That naturally covers rooted paths, .exe, and standalone binaries.

extensions/vscode-containers/src/runtimes/OrchestratorRuntimeManager.ts
bwateratmsft: Went with bubbling up an actionable error rather than auto-fixing. reconfigureClient now throws a clear message telling the user to set the base command only, and registration is wrapped in callWithTelemetryAndErrorHandling so it surfaces as a notification instead of breaking activation.

bwateratmsft: Decided not to auto-fix here. Compose V1 is long dead, so instead of guessing at a rewrite we throw a user-actionable error asking the user to use the base command.

bwateratmsft: Agreed. Chose to tell the user rather than own the risk of a bad rewrite. It now throws an actionable error pointing at the composeCommand setting.

bwateratmsft: Went this route. We no longer try to rewrite "docker-compose"; we throw an actionable error and ask the user to set the base command only.

patverb (approved):

PR #525, opened 2026-07-16
Restructure into pnpm monorepo (step 1)

bwateratmsft:
## Overview

Step 1 of the pnpm monorepo migration (#520): rearrange the **existing** repo into a pnpm workspace modeled on [`microsoft/vscode-azure`](https://github.com/microsoft/vscode-azure). This PR does **not** import the other repos into `packages/` — that is deferred to later PRs. Git history is preserved via `git mv` (`git log --follow` traces back through pre-move history).

## What changed

**Layout**
- Moved the entire extension into `extensions/vscode-containers/` (321 renames, history preserved).
- Deleted `package-lock.json`.

**New root scaffolding**
- `package.json` — private `@microsoft/vscode-containers-monorepo` with recursive `pnpm -r` scripts.
- `pnpm-workspace.yaml` — single workspace + catalog, overrides, `pnpmfile`, `trustLockfile`.
- `.npmrc` (internal feed proxy), `tsconfig.base.json`, `scripts/{postinstall,clean,pnpmfile}.mjs`, generated `pnpm-lock.yaml`.

**TypeScript 7 / tsgo**
- Root `@typescript/native` (typescript@7.0.2); `tsc --noEmit` runs tsgo.
- `js/ts.experimental.useTsgo` + recommended `typescriptteam.native-preview`.
- Extension's existing relaxed tsconfig flags preserved by re-parenting to `tsconfig.base.json`.

**Extension `package.json`**
- Azure-style scripts; shared build devDeps → `catalog:`; dropped root-owned devDeps.
- `homepage` → subdir README.
- Declared `@vscode/extension-telemetry` directly (the azext-eng telemetry bundle needs it under pnpm's strict `node_modules`).

**READMEs** — extension README moved into its subdir; new root README added (packages "coming soon", refs #520).

**Config** — `.vscode/{tasks,launch,settings,extensions}.json` updated for the subdir + `pnpm --filter vscode-containers`; `.gitignore`/`.gitattributes` updated; `.config/build.yml` → `packageManager: pnpm`; `.github/workflows/main.yml` → `package_manager: pnpm`. `.config/release.yml` unchanged (verified recursive vsix discovery + `Build Root` artifact still match).

## Verification

`pnpm install`, `pnpm build` (esbuild + tsgo type-check), `pnpm lint`, and `pnpm package` (valid 42-file vsix with bundles + LICENSE/NOTICE) all pass locally.

## Caveat to test

[...]

PR #530, opened 2026-07-17
Import compose-language-service into the monorepo

### LINKED ISSUE #520: Migrate to `pnpm` and monorepo
Right now, the Container Tools extensions and packages are split across four repos:

- microsoft/vscode-containers (this one)
- microsoft/vscode-docker
- microsoft/vscode-docker-extensibility
- microsoft/compose-language-service

With pnpm and a structure similar to the prototype Azure monorepo extension, we could put everything into this one repository, significantly reducing the burden of repo management. The rest can be archived.

bwateratmsft:
Imports [`microsoft/compose-language-service`](https://github.com/microsoft/compose-language-service) into this pnpm monorepo under `packages/compose-language-service/`, continuing the consolidation tracked in #520 (after #525 imported vscode-containers and #527 imported the vscode-docker pack).

## What this does

- **Preserves the full commit history**, rewritten so every path is nested under `packages/compose-language-service/` (`git filter-repo --to-subdirectory-filter`), replayed linearly onto the monorepo history — no graft/gut-point, single root, authorship/dates preserved.
- Wires `extensions/vscode-containers` to **build the library from its TypeScript source** (esbuild `alias` + tsconfig `paths` + `workspace:*`), mirroring `microsoft/vscode-azure`. The package keeps its own dual `tsc` build for the published npm artifact.
- Adds a **new npm-publish pipeline** (this repo currently only publishes VS Code extensions).
- `LICENSE.md` only (no `NOTICE.html`) — an npm library ships only its own source.

## Notes

- Merge with **Rebase and merge** to keep the rewritten history linear (temporarily enable "Allow rebase merging", then revert to squash-only). Do **not** squash or merge-commit.

## How to review the real changes

Because this PR imports the full commit history of `microsoft/compose-language-service`, the default "Files changed" tab renders every imported file as a brand-new add, which is not useful for review. The only changes that actually need review are the post-import integration commits.

Pristine import anchor: `e3e9a72` -- `fix(image-link): parse image refs explicitly and add GHCR/Quay support (#191)`. This is the last commit imported byte-for-byte from the source repo, immediately before the monorepo integration begins.

Integration commits (these are the real migration changes):

- `8de7fb8` Integrate compose-language-service into the monorepo (32 files, +103 / -4054)
- `ee20668` Add npm release pipeline and document the compose package (2 files, +4)
- `e35a06b` Apply pnpm-monorepo conventions to compose-language-service (33 files, +71 / -80)
- `b81b54a` Address PR review feedback: repository directory, hoist chai-as-promised, drop vscode from tsconfig types (2 files, +3 / -6)
- `bb74a7b` Stop committing the test client package-lock.json and gitignore it (2 files, +4 / -175)
- `598c040` Drop the redundant @microsoft/vscode-azext-eng devDep from compose (provided by the root) (1 file, -1)
- `bde5313` Fix lineAt bounds check and durations typing in compose telemetry (3 files, +31 / -40)
- `e403927` Address review: trim README packages note and revert tsconfig churn (2 files, +8 / -11)

Review links:

- Combined, in-PR "Files changed" view scoped to the post-import range (keeps review comments and per-file "Viewed" state): https://github.com/microsoft/vscode-containers/pull/530/files/e3e9a721f02e253447e963022af7acfa88b067bd..e403927879ed68487c83eb856f7e0d7923593698
- Per-commit diffs:
  - https://github.com/microsoft/vscode-containers/pull/530/commits/8de7fb8639820f965a8de2e18d77fe1bb5b8c173
  - https://github.com/microsoft/vscode-containers/pull/530/commits/ee20668738cbcb36ec71a1fa04a8cb2008c52afc
  - https://github.com/microsoft/vscode-containers/pull/530/commits/e35a06bfb8defa26ddce732d90732f1e0d5a42e8
  - https://github.com/microsoft/vscode-containers/pull/530/commits/b81b54ab39c420fa57866fe72f81dcc780b284bd
  - https://github.com/microsoft/vscode-containers/pull/530/commits/bb74a7b3e223ec7ae6c7225fe3f06e6c8d037b62
  - https://github.com/microsoft/vscode-containers/pull/530/commits/598c0403824182a998c759f72befefab3fb78938
  - https://github.com/microsoft/vscode-containers/pull/530/commits/bde5313a33ba80bc66d828c0a82e7ebea6a9cb96
  - https://github.com/microsoft/vscode-containers/pull/530/commits/e403927879ed68487c83eb856f7e0d7923593698

Note: if more commits are pushed, bump the end SHA in the range link to the new HEAD, or just shift-select the post-anchor commits on the Commits tab to regenerate the scoped view.

Draft while integration + pipeline work and the writeup/`npm pack`-diff deliverables are finalized.


bwateratmsft:
## Deliverable 1 — Per-file writeup (everything not byte-identical to `microsoft/compose-language-service`)

The history import (`git filter-repo --to-subdirectory-filter`) landed the source tree **byte-identical** under `packages/compose-language-service/` at commit `e3e9a721`. Everything below is a follow-up change on top of that byte-identical baseline.

### A. Files removed from the package (root already provides them)

Deleted as part of the include/exclude matrix — the monorepo root owns these, so a per-package copy would be redundant or wrong:

| Removed | Why |
| --- | --- |
| `.github/**` (CODEOWNERS, 3 workflows) | Root owns CI, issue automation, and CODEOWNERS. |
| `.config/{build,release,tsaoptions}.yml/json` | Pipelines live at the repo root (see Deliverable 3). |
| `.vscode/**` (extensions, launch, settings, tasks) | Root owns the workspace `.vscode`; no sibling package/extension keeps its own. |
| `LICENSE`, `NOTICE.html` | `LICENSE.md` is supplied per-package by `scripts/postinstall.mjs`; an npm library ships no `NOTICE.html`. |
| `SECURITY.md` | Root owns it. |
| `package-lock.json` | Monorepo uses pnpm + the root lockfile. |
| `.gitignore`, `.nvmrc` | Root versions win. |
| `.gitattributes` | Folded into the root `.gitattributes` (see below). |

### B. Package files edited

**`package.json`**
- `repository.url` / `bugs.url` / `homepage` → repointed to `microsoft/vscode-containers` (homepage now deep-links to `packages/compose-language-service/README.md`).
- `scripts`: `npm` → `pnpm` for `build`/`package`; added `clean` → `../../scripts/clean.mjs` (matches siblings).
- `exports`: reordered so the `types` condition is **first** in each of the `.`/`./client`/`./vscode` blocks (condition order matters — this also clears esbuild's "types resolved after import/require" warnings).
- Dependencies retargeted to the LSP **3.18 stable** line already used elsewhere in this repo (per maintainer direction): `vscode-languageserver ^10.1.0`, `vscode-languageserver-protocol ^3.18.2`; devDeps `vscode-languageclient` (`catalog:`, resolving to `^10.1.0`), `vscode-jsonrpc ^9.0.1`, `vscode-languageserver-types ^3.18.0`. Aligned `@microsoft/vscode-azext-eng` to `^1.1.0-alpha.3`. (`chai-as-promised`/`@types/chai-as-promised` are hoisted at the root rather than declared here; see section E.)
- Removed `@types/node` and `@types/vscode` devDeps — these are provided by the monorepo and pulled in via the tsconfig `types` field instead (see below).

**`tsconfig.json`**
- Added `"types": ["node", "mocha"]` (originally also `"vscode"`, dropped in review; see section E). The native TS compiler (tsgo) does not auto-include ambient `@types/*` globals from a nested package under pnpm, so they're named explicitly. This is what replaces the removed `@types/node` dep; `@types/vscode` (also provided by the monorepo) is resolved through the explicit `vscode` module imports in `src/vscode`.

**Source — adjustments for the LSP 3.18 API drift** (the upgraded protocol tightened several signatures):
- `src/service/ComposeLanguageService.ts` — bridge `createLspHandler`'s generics to 3.18's stricter `ServerRequestHandler<P,R,PR,E>`: infer `R`/`E` from the connection side via `NoInfer<>` (keeps provider validation) and use `any` for the partial-result type `PR` (the connection types it as the item type while providers use `never`; these are contravariantly irreconcilable even though the wrapper forwards the reporter unchanged). `callWithTelemetryAndErrorHandling`'s callback return type widened to `Promise<R | ResponseError<E>>` so the wrapper matches the `Thenable` branch of `HandlerResult` with no cast.
- `src/service/ComposeDocument.ts` — `let indentDepth: number;` (was seeded with an unused value; the monorepo eslint flags `no-useless-assignment`).
- `src/vscode/DocumentSettingsClientFeature.ts` — dropped an unnecessary `as DocumentSettings` assertion.
- `src/test/providers/DiagnosticProvider.test.ts` — `Diagnostic.message` is now `string | MarkupContent` in 3.18; narrow before `.includes`.
- `src/test/utils/telemetry/TelemetryAggregator.test.ts` — removed a now-unused `eslint-disable`.

### C. Monorepo-level files touched (outside the package)

- `pnpm-workspace.yaml` — add `packages/compose-language-service`.
- `scripts/postinstall.mjs` — copy `LICENSE.md` into the package (LICENSE-only).
- `.gitattributes` — force LF on `packages/compose-language-service/bin/docker-compose-langserver`.
- `extensions/vscode-containers/package.json` — `@microsoft/compose-language-service`: `^0.5.0` → `workspace:*`.
- `extensions/vscode-containers/esbuild.mjs` — alias the `/client` and `/vscode` subpaths to the package **source**, and repoint the compose **server** esbuild entry from `node_modules/.../dist/esm/server.js` to `../../packages/compose-language-service/src/server.ts` (build-from-source; out path unchanged).
- `extensions/vscode-containers/tsconfig.json` — `rootDir` `src` → `../..` and `paths` for the two subpaths, so `tsc --noEmit` and eslint resolve the same source esbuild bundles.
- `README.md` — Packages section entry for the library.
- `pnpm-lock.yaml` — regenerated by the install.

### D. Follow-up: pnpm-monorepo conventions (commit `e35a06b`)

A later pass aligned compose with the same conventions applied to the other imported package (PR #531). None of this changes the published artifact's file set (see Deliverable 2):

- **tsconfig dedupe.** Shared compiler options (`declaration`, `stripInternal`, `types: ["node"]`, alongside the existing module/target/strict block) now live in the root `tsconfig.base.json`. `packages/compose-language-service/tsconfig.json` was slimmed to `extends: ../../tsconfig.base.json` plus only `outDir`, `rootDir`, `inlineSources`, and its own `types: ["node", "mocha"]` (types arrays do not merge across `extends`, so `node` is repeated). `extensions/vscode-containers/tsconfig.json` gained `declaration: false`, because the esbuild-bundled app type-checks with `tsc --noEmit` and would otherwise fail TS2883 from inheriting the base `declaration: true`.
- **Catalog.** `vscode-languageclient` (shared by the extension and compose) moved to the pnpm `catalog:` (entry `^10.1.0` in `pnpm-workspace.yaml`); both consumers now reference `catalog:`. On `pnpm pack` this resolves back to `^10.1.0`, so the published package.json still pins a concrete range.
- **Mocha loader.** Compose's test loader moved off `tsx` to Node native type-stripping: `node-option` is now `["experimental-transform-types", "import=@microsoft/vscode-azext-eng/mocha"]`, and the `tsx` devDep was dropped. Because `--experimental-transform-types` strips type annotations but does NOT elide type-only imports, `import type` / `export type` discipline was enforced across the package (about 30 `src` files) and the `chai/register-should` side-effect import gained an explicit `.js` so Node ESM resolves it. Emitted output and behavior are unchanged; all 128 tests still pass.

The package still keeps its own dual `tsc` build + `exports` → `dist`, so the published npm artifact is unchanged in shape (see Deliverable 2).

### E. Review follow-ups (commit `b81b54a`)

Small refinements from PR review, none of which change the published artifact's file set (Deliverable 2):

- **`repository.directory`.** Added `"directory": "packages/compose-language-service"` to the package.json `repository` block so the npm metadata points at the package's location within the monorepo (matches the convention used in #531).
- **Hoisted test deps.** `chai-as-promised` and `@types/chai-as-promised` are now hoisted at the root `publicHoistPattern` (both are provided transitively by `@microsoft/vscode-azext-eng`) instead of being declared in the compose `devDependencies`, mirroring how `chai`/`@types/chai`/`mocha` are already handled.
- **tsconfig `types`.** Dropped `vscode` from the compose tsconfig `types` array (now `["node", "mocha"]`); no source changes were needed, since the files under `src/vscode` already import the `vscode` module explicitly.

### F. Review follow-ups (commit `bde5313`)

Two correctness fixes from Copilot review, neither of which changes the published artifact's file set (Deliverable 2):

- **`ComposeDocument.lineAt` bounds.** Tightened the out-of-bounds guard from `> lineCount` to `>= lineCount`, since valid line indices are `0..lineCount-1`; the old `>` let `line === lineCount` through.
- **`TelemetryAggregator` durations typing.** Switched the `durations` filter to a type guard (`(d): d is number`) so it is a proper `number[]` for `logNormal(values: number[])`, dropping a redundant `?? undefined` and `|| []`.

bwateratmsft:
## Deliverable 2 — `npm pack` diff (published `@microsoft/compose-language-service@0.5.0` vs. this branch)

Compared the currently-published tarball against a fresh `pnpm pack` from this branch.

|  | Published 0.5.0 | This branch |
| --- | --- | --- |
| Tarball size | 36.0 KB | 35.7 KB |
| File count | 103 | 101 |

### File-set differences (the *only* structural changes)

**Removed (2):**
- `NOTICE.html` — an npm library ships no third-party NOTICE.
- `SECURITY.md` — owned by the monorepo root, not shipped in the library.

**Renamed (1):**
- `LICENSE` → `LICENSE.md` — the monorepo supplies `LICENSE.md` per-package via `scripts/postinstall.mjs`; contents are the same MIT license text.

Everything else is present in **both** tarballs and in the same layout:
- `dist/esm/**` (JS + `.d.ts`) and `dist/cjs/**` (JS only — `build:cjs` runs with `--declaration false`),
- `bin/docker-compose-langserver`,
- `README.md`, `CHANGELOG.md`, `package.json`.

`.npmignore` (kept from the source repo) still excludes `dist/*/test/**`, `src`, `tsconfig.json`, `.config`, `*.map`, etc., so the built `dist/esm/test/**` output does **not** leak into the tarball.

### Content differences (same file names, changed bytes)

- `package.json` — repointed URLs, `pnpm` scripts, `types`-first `exports`, the LSP 3.18 dependency ranges, and the migrated mocha loader config (`experimental-transform-types` + `@microsoft/vscode-azext-eng/mocha`). The `vscode-languageclient` `catalog:` reference resolves back to `^10.1.0` on pack, so the shipped file still pins a concrete range (Deliverable 1B/1D). It also adds `repository.directory` and no longer lists the `chai-as-promised`/`@types/chai-as-promised` devDeps, which are now hoisted at the root (Deliverable 1E).
- `dist/**/*.js` + `.d.ts` — regenerated from source built against the LSP 3.18 stable stack (plus the small source fixes in Deliverable 1B). Functionally equivalent; byte differences are the expected result of the dependency bump + recompile.

**Net:** the publishable artifact is structurally identical apart from dropping `NOTICE.html`/`SECURITY.md` and the `LICENSE` → `LICENSE.md` rename — all deliberate and consistent with how the other packages in this monorepo are published.

bwateratmsft:
## Deliverable 3 — Pipeline evaluation (build + publish)

### (a) Build — existing `.config/build.yml` reused as-is

No pipeline change was needed. `.config/build.yml` already extends the shared `1es-mb-main.yml` template with `packageManager: pnpm`, which runs the recursive `build` / `lint` / `test` / `package` scripts. Because the package is now in `pnpm-workspace.yaml`, that recursive pass automatically:

- builds compose (dual esm/cjs `tsc`), lints it (`eslint --max-warnings 0`), and runs its 128 mocha tests;
- runs `pnpm pack` for compose, emitting **`microsoft-compose-language-service-0.5.0.tgz`** into the build artifact alongside the extension VSIXes;
- builds the `vscode-containers` extension, whose esbuild config now **bundles the compose language server from source** — verified locally: `dist/compose-language-service/lib/server.js` (349 KB) contains the compose source classes (`ComposeLanguageService`, providers, …), not a copy pulled from `node_modules`.

Local verification run: `pnpm -r run build` ✅, `pnpm -r run lint` ✅, compose `test` ✅ (128 passing), extension `build:check` (`tsc --noEmit`) ✅, `pnpm --filter @microsoft/compose-language-service pack` ✅.

### (b) Publish — new `.config/release-npm.yml`

The repo previously only published VS Code extensions (`.config/release.yml` → `1es-mb-release-extension.yml`, VSIX). Publishing an npm library needs a separate pipeline, added as **`.config/release-npm.yml`** (the extension release pipeline is untouched):

- `trigger: none`, `pr: none` — manually triggered, with `packageToPublish` (`microsoft-compose-language-service`), `publishVersion`, and `dryRun` parameters.
- `resources.pipelines.build.source` → the monorepo build pipeline (`\Azure Tools\VSCode\Extensions\vscode-containers`), so it consumes the `*.tgz` produced in (a).
- `extends: azdo-pipelines/1es-mb-release-npm.yml@azExtTemplates` with `ownerAliases`, `approverAliases`, `gitHubServiceConnection`, and `feedBaseUrl` wired from the shared `azcode.variables.yml`, plus `releaseApprovalEnvironment: VSCodeDockerExtensionPublish`.

This mirrors the source repo's own `.config/release.yml` (which already used this npm template), just repointed at the monorepo build pipeline. The template handles ESRP publish to npmjs.org under the `@microsoft` scope, version gating (package.json version must equal `publishVersion`), artifact selection by the `<name>*.tgz` prefix, and the optional Azure Artifacts feed mirror via `feedBaseUrl`.

**Open item — provenance/OIDC:** publishing goes through ESRP rather than `npm publish --provenance`, so npm provenance attestation is not automatically produced. If provenance is required for this package, it needs to be confirmed/added at the template level; flagging rather than assuming.

packages/compose-language-service/package.json
```diff
@@ -7,12 +7,12 @@
     "license": "MIT",
     "repository": {
         "type": "git",
-        "url": "https://github.com/microsoft/compose-language-service"
+        "url": "git+https://github.com/microsoft/vscode-containers.git"
```
bwateratmsft: Can we do `directory` here, like in #531?

packages/compose-language-service/package.json
```diff
@@ -21,50 +21,54 @@
...
         "node": ">=22"
     },
     "devDependencies": {
-        "@microsoft/vscode-azext-eng": "1.0.0-alpha.12",
-        "@types/node": "22.x",
-        "@types/vscode": "1.105.0",
-        "vscode-languageclient": "^10.0.0-next.19"
+        "@microsoft/vscode-azext-eng": "^1.1.0-alpha.3",
+        "@types/chai-as-promised": "8.0.2",
+        "chai-as-promised": "8.0.2",
```
bwateratmsft: These can be hoisted at the root so they can be removed here.

packages/compose-language-service/package.json
```diff
@@ -21,50 +21,54 @@
...
+        "vscode-jsonrpc": "^9.0.1",
+        "vscode-languageclient": "catalog:",
+        "vscode-languageserver-types": "^3.18.0"
```
bwateratmsft, line 58: If this is used in any other packages or the extension, let's do a catalog version. Same for the other `vscode-languageserver*` packages and `vscode-languageclient` too.

packages/compose-language-service/tsconfig.json
```diff
@@ -1,22 +1,17 @@
...
-        "strict": true,
-        "stripInternal": true,
+        "inlineSources": false,
+        // `types` arrays do not merge across `extends`; repeat `node` (from the base)
+        // alongside the extra ambient types this package needs (mocha for its tests,
+        // vscode for the client-side features under src/vscode).
+        "types": [
+            "node",
+            "mocha",
+            "vscode",
```
bwateratmsft: Let's remove `vscode` from here and make source changes if needed to fix compile

bwateratmsft: Accepted. I removed vscode from the types array and confirmed the compiled src/vscode usages already import the vscode module explicitly, so no source changes were needed.

packages/compose-language-service/package.json
bwateratmsft: Accepted. I added the repository directory field pointing at packages/compose-language-service so npm metadata identifies the package location in the monorepo.

packages/compose-language-service/package.json
bwateratmsft: Accepted. I verified chai-as-promised and @types/chai-as-promised are already provided by @microsoft/vscode-azext-eng at 8.0.2, then hoisted them in pnpm-workspace.yaml and removed the duplicate devDependencies here.

packages/compose-language-service/package.json
bwateratmsft: Partially accepted. vscode-languageclient is already shared and already uses catalog:, while the server-side vscode-languageserver*, vscode-jsonrpc packages only appear in compose-language-service, so I left those pinned locally.

packages/compose-language-service/src/service/ComposeDocument.ts
```diff
@@ -0,0 +1,356 @@
...
+        // Flatten to a position at the start of the line
+        const startOfLine = (typeof line === 'number') ? Position.create(line, 0) : Position.create(line.line, 0);
+        const endOfLine = Position.create(startOfLine.line, MaximumLineLength);
+
+        if (startOfLine.line > this.textDocument.lineCount) {
+            throw new Error(`Requested line ${startOfLine.line} is out of bounds.`);
+        }
+
+        return this.textDocument.getText(Range.create(startOfLine, endOfLine));
+    }
```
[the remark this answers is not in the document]
bwateratmsft: Good catch, fixed. Changed the bounds check from `>` to `>=` so a line index equal to lineCount (which is out of range, since valid indices are 0..lineCount-1) is now correctly rejected.

packages/compose-language-service/src/service/utils/telemetry/TelemetryAggregator.ts
```diff
@@ -0,0 +1,113 @@
...
+        // For each group, aggregate properties and add performance statistics, to get one aggregated event per group
+        for (const events of eventGroups.values()) {
+            const eventName = events[0].eventName;
+
+            const aggregatedEvent = initEvent(eventName);
+
+            // Aggregate the performance statistics
+            const durations = events.map(e => e.measurements.duration ?? undefined).filter(d => d !== undefined) || [];
+            const stats = logNormal(durations);
+
```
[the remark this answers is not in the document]
bwateratmsft: Fixed. Switched to a type-guard filter `(d): d is number => d !== undefined` so durations is a proper number[], and dropped the redundant `?? undefined` and `|| []` since duration is already number | undefined and map/filter always returns an array.

patverb (approved):

PR #561, opened 2026-07-27
Add wslc support

patverb:
Add  wslc  (WSL Container CLI) as a Windows-only container runtime (preview)
Description

Registers the new  WslcClient  from  @microsoft/vscode-container-client  as a selectable, Windows-only container runtime, exposed as "WSLC (Windows only, preview)". It's treated as an alternative to Docker/Podman, mirroring the existing auto-configurable client pattern.

Depends on microsoft/vscode-docker-extensibility# ( @microsoft/vscode-container-client@^0.6.0 ).

What's added / changed
•  AutoConfigurableWslcClient  — mirrors the Docker/Podman auto-configurable clients, but pins the command to  wslc  (ignores the shared  containers.containerCommand  override, since wslc is a specific named binary).
•  extension.ts  — registers the wslc client only on Windows ( isWindows()  gate), and wires up  reconfigure .
•  RuntimeManager  — if  wslc  is selected on a non-Windows host, rejects with a clear "WSLC is only available on Windows" error instead of a confusing timeout.

.NET SDK container task support
The .NET SDK's  LocalRegistry  publish target only supports Docker/Podman. To support wslc, the SDK run task now:

Publishes the image to a temporary tar archive ( -p:ContainerArchiveOutputPath=<image>-.tar ) instead of  -p:LocalRegistry ,
Loads it with  wslc load --input  ,
Runs it — then deletes the temp tar in a  finally  (GUID-named to avoid collisions).
Docker/Podman continue to use  LocalRegistry  unchanged.

Fixes #501

PR #565, opened 2026-07-28
vscode-container-client: dedupe runtime clients with shared Zod records and helpers

bwateratmsft:
## TODO
- [x] Docker Desktop integration tests pass
- [x] Podman integration tests pass
- [x] Nerdctl integration tests pass

## Summary

Reduces duplication and near-duplication in `packages/vscode-container-client` (heaviest in `src/clients`), most of which stemmed from small per-runtime differences in the shapes of objects emitted by Docker, Podman, and Nerdctl. Zod transforms let us normalize those differences in one place, shared base helpers let the clients stop re-implementing the same parse/wrapper boilerplate, and the test suite is realigned so each unit test sits next to (and is named after) its source.

**Net: ~930 lines of production code removed** (+1320 / -2249, excluding `src/test/`). Counting tests too, it's 60 files changed with rename detection, +1692 / -2384.

### 1. Consolidated per-runtime record schemas → `Shared*Record.ts` + Zod transforms
The per-runtime `*Record.ts` schemas (list + inspect for images, containers, networks, volumes across Docker/Podman/Nerdctl) were near-identical. They're replaced with a single set of `Shared*Record.ts` schemas under `DockerClientBase/`, using Zod transforms (`ZodTransforms.ts`) to normalize each runtime's small shape differences (label maps, OS/arch aliases, sizes, image names, scalar-vs-array fields, absent `size`/`createdAt`, etc.). Runtimes that emit a JSON array (Podman) vs newline-delimited JSON (Docker/Nerdctl) are unified at the parse layer, so a single tolerant schema serves all three.

Where the runtimes genuinely disagree, the shared normalizers take a small options object, and each runtime's settings are bundled into a named constant (`DockerListContainerOptions` / `NerdctlListContainerOptions`, `DockerListImageOptions` / `NerdctlListImageOptions`) so call sites stay declarative and invalid combinations aren't spelled out ad hoc.

### 2. Routed base parsing through shared helpers, deleted redundant overrides
The default `parse*CommandOutput` methods in `DockerClientBase` now go through shared `parsePerLineJson` / `parseInspectJson` helpers instead of hand-rolled loops. Client overrides that had become byte-for-byte identical to the base were deleted (Podman/Nerdctl inspect + list variants, including the Podman and Nerdctl `parseListVolumes` overrides). `parseInspectJson` unifies the "single JSON array" and "per-line JSON" paths, so Podman inspect now also tolerates per-item errors in non-strict mode (matching Docker/Nerdctl).

### 3. Prune helper + command-response factory on the shared ancestor
- Added `resolvePrunedIds` for runtimes that emit only deleted IDs on prune.
- Added a command-response factory (`makeCommandResponse` / `makeVoidCommandResponse` / `makeStreamCommandResponse`) on the shared `ConfigurableClient` ancestor, collapsing ~50 repetitive `Promise.resolve({ command, args, parse })` wrapper methods across **both** the docker and compose client hierarchies into single calls.

### 4. Smaller opportunistic cleanups
- **Podman volume unification:** all three runtimes now share one tolerant list-volume schema/normalizer; the Podman/Nerdctl overrides and the redundant `DockerVolumeRecord.ts` are gone.
- **Dropped Podman v3 network tolerance** (v3/v4 are EOL): removed v3-only `Name`/`Labels` handling and fixed a latent bug where v4+ lowercase `labels` were silently dropped from `network ls`.
- **Hoisted `resolveCreatedAt`** into its own module with a `resolveCreatedAtBaseline` helper, reused by the inspect normalizer instead of duplicating the parse-with-now-fallback logic. Its mode is a `CreatedAtMode` union (`'lenient' | 'validated'`) rather than a bare boolean.
- **Removed `parseDockerLikeLabels`**, a wrapper whose last production caller disappeared with the shared volume schema; its cases moved onto `labelsStringSchema` in `ZodTransforms.test.ts`.

### 5. Realigned the unit test layout
Each unit test now mirrors its source file name (`foo.ts` → `foo.test.ts`) and lives under a `src/test/` subtree that matches the source's directory (`clients/DockerClientBase/`, `contracts/`, `utils/`, …). Tests that had been individualized against one source were merged back onto that source's name (e.g. `normalizeContainerState.test.ts` → `SharedListContainerRecord.test.ts`, `NerdctlInspectContainerRecord.test.ts` → `SharedInspectContainerRecord.test.ts`).

### 6. Integration test harness
Needed to actually run the Podman and Nerdctl integration suites against the consolidated schemas:
- Added devcontainer setup files for the Podman and Nerdctl integration environments (`src/test/podman/`, `src/test/nerdctl/`).
- Added a `pretestintegration` script so `@microsoft/vscode-processutils` is built before the integration run.
- Compare bind-mount sources through `wslifyPath` in the E2E test: on Windows, machine-based runtimes report the source as a translated `/mnt/<drive>` path rather than the original `D:\...` host path.
- Normalize generated compose container names before comparison, since podman-compose uses `_` separators where Docker/Nerdctl use `-`.

## Behavior notes for reviewers
- **Non-strict resilience (intentional):** Podman inspect previously failed the whole batch if a single item was malformed even in non-strict mode. It now swallows per-item errors in non-strict mode, matching Docker/Nerdctl.
- **Raw text fix:** each inspect item now carries its own per-item JSON as `raw`. Previously Podman handed consumers the entire command output, so `JSON.parse(result[0].raw)` in the extension's inspect commands got the whole array.
- **Container timestamp baseline fix:** the inspect normalizer now compares `startedAt`/`finishedAt` against a validated created-at baseline, so an invalid `Created` no longer wrongly includes past start/finish times.
- **OS/architecture fix:** the old Docker/Podman inspect-image normalizers tested `Architecture === 'windows'` (instead of `Os`) when detecting the operating system, and did not recognize the `x86_64`/`aarch64` aliases. Both are fixed by the shared transforms.
- **Network booleans are tolerant:** `IPv6`/`Internal` in `network ls` fall back to `false` on an empty or unrecognized value rather than failing the whole record, matching the previous `=== 'true'` comparison.
- **Docker `<none>` tags are preserved:** dropping a `<none>` tag from the composed image name is a Nerdctl-only behavior (`NerdctlListImageOptions`), so Docker's `originalName` is unchanged.
- **Left intentionally hand-rolled:** Podman `parseListImages` (stateful tag indexing across items) can't be expressed via the generic helper and was kept as-is.

## Validation
- Build clean (`vscode-processutils` + `vscode-container-client`)
- 126 unit tests passing
- Lint: 0 warnings

## Note
This overlaps substantially with the schema work in #561 — a chunk of the lines that PR adds are made redundant here.

Draft for early review; happy to split into reviewable commits if preferred.


packages/vscode-container-client/src/clients/DockerClientBase/DockerClientBase.ts
```diff
@@ -1222,55 +1097,35 @@ export abstract class DockerClientBase extends ConfigurableClient implements ICo
...
+
+            return {
+                name: rawVolume.Name,
+                driver: rawVolume.Driver,
+                labels,
+                mountpoint: rawVolume.Mountpoint,
+                scope: rawVolume.Scope,
+                createdAt: createdAt?.toDate(),
+                size
+            };
```
bwateratmsft: Why does this look so different from the other method bodies in here?

bwateratmsft: The normalizer is just inline, that's why

packages/vscode-container-client/src/clients/DockerClientBase/SharedListContainerRecord.ts
```diff
@@ -0,0 +1,241 @@
...
+        }
+        return parsed.toDate(); // Docker: preserve the (invalid) parse result
+    }
+
+    if (validateDate && strict) {
+        throw new Error('Container creation date is missing');
+    }
+    // Docker: dayjs.utc(undefined) yields the current time
+    return validateDate ? new Date() : dayjs.utc(raw).toDate();
+}
```
bwateratmsft: This looks reusable elsewhere

packages/vscode-container-client/src/test/clients/DockerClientBase/SharedInspectContainerRecord.test.ts
bwateratmsft, line 1: This file should be renamed/merged

bwateratmsft: Done in 3b712161. Renamed to `SharedInspectContainerRecord.test.ts` (it tests `SharedInspectContainerRecord.ts`, not nerdctl) and applied the convention across the suite: each unit test now mirrors its source file name and lives under a `src/test/` subtree matching the source's directory (e.g. `clients/DockerClientBase/`, `contracts/`, `utils/`). The individualized tests that mapped to a single source were merged onto that source's name (`normalizeContainerState.test.ts` -> `SharedListContainerRecord.test.ts`, `DockerClientBaseParsing.test.ts` -> `DockerClientBase.test.ts`).

packages/vscode-container-client/src/clients/DockerClientBase/SharedListContainerRecord.ts
bwateratmsft: Done in 3b712161. Extracted `resolveCreatedAt` into its own module (`resolveCreatedAt.ts`) and added a `resolveCreatedAtBaseline` helper. The inspect normalizer now reuses that baseline (parse-with-now-fallback) instead of duplicating the logic inline.

packages/vscode-container-client/src/contracts/ZodTransforms.ts
```diff
@@ -115,3 +115,37 @@ export const architectureStringSchema = z.pipe(z.string(), z.transform((str): 'a
...
+    z.nullish(z.union([z.string(), z.number()])),
+    z.transform((val): number | undefined => tryParseSize(val ?? undefined)),
+);
+
+/**
+ * Schema that transforms a raw image name string into an {@link ImageNameInfo}
+ * via {@link parseDockerLikeImageName}. A `null`/`undefined`/empty value yields
+ * an {@link ImageNameInfo} with only `originalName` populated.
+ */
+export const imageNameSchema = z.pipe(
```
patverb, line 136: Is `imageNameSchema` or `stringArraySchema` used anywhere?

packages/vscode-container-client/src/clients/DockerClientBase/SharedListImageRecord.ts
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * A single, tolerant schema for the Docker-compatible `image ls` per-line JSON
+ * emitted by Docker and nerdctl. Docker emits `ID`/`Size` unconditionally;
+ * nerdctl may omit them. The size and creation date are normalized by the shared
+ * field transforms.
+ *
+ * Podman's `image ls` output is object-shaped (`Names` array, numeric `Created`)
+ * and keeps its own record module rather than sharing this schema.
+ */
+export const SharedListImageRecordSchema = z.object({
```
patverb, line 20: We have tests for the `SharedListConntainerRecord` should we have them for this too? Same is true for alot of these new "Shared...Records", should we unit test those too? We do test all this stuff in the integration tests which tbh is more valuable.

packages/vscode-container-client/src/clients/DockerClientBase/SharedListImageRecord.ts
```diff
@@ -0,0 +1,75 @@
...
+    ID: z.optional(z.string()),
+    Repository: z.string(),
+    Tag: z.optional(z.string()),
+    // Date string transformed to Date with fallback to current time
+    CreatedAt: z.optional(dateStringWithFallbackSchema),
+    // Size (bytes number or human-readable string) transformed to number | undefined
+    Size: z.optional(sizeSchema),
```
patverb: Isnt sizeSchema already optional? you didnt do `z.optional()` in `SharedListVolumeRecord`

packages/vscode-container-client/src/clients/DockerClientBase/SharedInspectImageRecord.ts
```diff
@@ -1,95 +1,105 @@
...
+ *
+ * `Architecture`, `Os`, and `Created` reuse the shared field transforms so the
+ * schema does the normalization once for every runtime. This also corrects a
+ * latent bug in the old Docker/Podman normalizers, which tested
+ * `Architecture === 'windows'` (instead of `Os`) when detecting the operating
+ * system, and did not recognize the `x86_64`/`aarch64` architecture aliases.
+ */
+
+const InspectImageConfigSchema = z.object({
+    Entrypoint: z.optional(z.union([z.array(z.string()), z.string(), z.null()])),
```
patverb: Should these be the `stringArraySchema` from the zod transformers? I dont see them used anywhere.

packages/vscode-container-client/src/clients/DockerClientBase/SharedInspectContainerRecord.ts
```diff
@@ -0,0 +1,212 @@
...
+
+type InspectContainerMount = z.infer<typeof InspectContainerMountSchema>;
+
+const InspectContainerNetworkSchema = z.object({
+    Gateway: z.optional(z.string()),
+    IPAddress: z.optional(z.string()),
+    MacAddress: z.optional(z.string()),
+});
+
+const InspectContainerConfigSchema = z.object({
```
patverb, line 59: General Observation: Even though strict is getting passed in for docker, most things are now optional.

patverb (dismissed):

packages/vscode-container-client/src/clients/DockerClientBase/SharedInspectImageRecord.ts
bwateratmsft: Yes — done in 098064b2. `Entrypoint`/`Cmd` here (and in `SharedInspectContainerRecord`) now use `stringArraySchema`, so the string-or-array coalescing happens in the schema and the inline `toArray(... || [])` calls in the normalizers are gone.

packages/vscode-container-client/src/clients/DockerClientBase/SharedInspectContainerRecord.ts
bwateratmsft: Right — that's an intentional tradeoff of the consolidation. A single tolerant "superset" schema validates all three runtimes' output, so Docker's stricter shape is modeled as a subset (everything any runtime may omit is optional/nullable). We lose the tight per-field validation that a Docker-only schema would give, but that's what lets the three runtimes share one schema + normalizer. The real per-runtime shapes are exercised by the integration tests. If we ever want stronger Docker validation we could layer a Docker-specific refinement on top, but I'd rather not reintroduce the duplication we just removed unless we hit an actual need.

packages/vscode-container-client/src/clients/DockerClientBase/SharedListImageRecord.ts
bwateratmsft: You're right — `sizeSchema` is fronted by `z.nullish(...)`, which already makes the object key optional (verified: `z.object({ Size: sizeSchema }).parse({})` succeeds). The outer `z.optional()` was redundant, so I removed it here and in `SharedListVolumeRecord` in 098064b2. (The date schemas are `z.pipe(z.string(), …)` and still need `z.optional()`, so those stay.)

packages/vscode-container-client/src/clients/DockerClientBase/SharedListImageRecord.ts
bwateratmsft: Agreed the integration tests are the higher-value coverage here. That said, I added focused unit tests in 098064b2 for every `Shared*Record` normalizer (`SharedListImageRecord`, `SharedListVolumeRecord`, `SharedListNetworkRecord`, `SharedInspectImageRecord`, `SharedInspectVolumeRecord`, `SharedInspectNetworkRecord`) plus `normalizeListContainerRecord`, concentrating on the branching logic (option knobs, `<none>`-tag handling, `isLocalImage`, Config-vs-top-level fallbacks, IPAM filtering, port-skip vs throw). +30 tests.

packages/vscode-container-client/src/contracts/ZodTransforms.ts
bwateratmsft: Good catch — they weren't used anywhere yet. Fixed in 098064b2: `stringArraySchema` now backs the `Entrypoint`/`Cmd` fields in both `SharedInspectImageRecord`/`SharedInspectContainerRecord` (see your other comment), and `imageNameSchema` now backs the `Image` field in `SharedListContainerRecord`, so both are genuinely wired in.

patverb (approved):

PR #568, opened 2026-07-31
vscode-container-client: add FinchClient and FinchComposeClient

bwateratmsft:
Adds first-class `FinchClient` and `FinchComposeClient` for [Finch](https://github.com/runfinch/finch), and registers them.

Finch is nerdctl + containerd + BuildKit inside a Lima VM (WSL2 on Windows), and forwards nearly every subcommand to nerdctl verbatim, so both new clients subclass their nerdctl counterparts and emit identical command lines.

## The blocker: Finch couldn't be registered at all

`NerdctlClient`'s constructor passed a hard reference to its own `static ClientId` to `super(...)`, so a subclass that merely redeclares `static ClientId` would still register under the nerdctl ID, and `RuntimeManager.registerRuntimeClient` throws on duplicate IDs.

Fixed by adding an optional trailing `clientId` parameter to the `NerdctlClient` and `NerdctlComposeClient` constructors, defaulting to their own IDs. Existing signatures and callers are unaffected. This is the only behavioural change to `NerdctlClient`.

There's a unit test asserting `new FinchClient().id === FinchClient.ClientId` and `!== NerdctlClient.ClientId` (plus the inverse for nerdctl), since this is easy to silently get wrong.

Worth noting given #561: `officialRuntimeRegistrations` uses `ClientConstructor<T> = (new () => T) & { readonly ClientId: string }` and reads `client.ClientId` for telemetry, so a subclass that inherited its parent's static would have been misreported there too. The constructor parameter makes the static and the instance ID agree.

## Version reporting

`finch version` is one of the few commands Finch implements itself rather than forwarding. It nests the nerdctl version under `Client.NerdctlClient` and reports the Finch version as `Client.Version`. Captured from a real Finch v1.17.2 install:

```jsonc
{
  "Client": {
    "Version": "v1.17.2",                                        // Finch
    "NerdctlClient": { "Version": "v2.2.2", "Os": "linux", ... } // nerdctl
  },
  "Server": { "Components": [ { "Name": "containerd", "Version": "v2.2.1" }, ... ] }
}
```

An earlier revision of this PR treated the inherited nerdctl parsing as a bug and overrode it to dig out the nested nerdctl version. Review feedback pushed back on that, and the pushback was right: `VersionItem.client` is reported by every other client as the version of *the CLI you invoke* (Docker reports docker, Podman podman, wslc wslc, nerdctl nerdctl), and `checkInstall` runs `finch -v`, which prints the Finch version. Reporting nerdctl would have made `version()` disagree with `checkInstall()` and broken with every sibling client.

So **`client` is the Finch version**, which is exactly what the inherited nerdctl parsing already produces — `Client.Version` — and `Server` is passed through by Finch unmodified, so the containerd component lookup works as-is. The override and its Zod schema were redundant and have been removed.

**This means there is no version-parsing fix in this PR.** The pre-existing behavior was already correct; it just wasn't intentional or tested. The unit tests are kept precisely because Finch's shape differs from nerdctl's, so it's worth pinning that `client` is `v1.17.2` and not the nested `v2.2.2`, with a comment on the class documenting the shape so nobody "fixes" it back.

Verified against the live CLI, the two now agree:

```
version():      {"client":"v1.17.2","server":"v2.2.1"}
checkInstall(): "finch version v1.17.2"
```

The nerdctl version is deliberately not surfaced. It's arguably the more functionally meaningful number (nerdctl determines which flags actually work), but there's no field for it and `client` semantically means the CLI being invoked, so it's treated as a VM implementation detail — like containerd's `runc` component, which we also don't surface.

## Wiring

Thanks to #561's centralized registration table, the extension-side change is one line:

```ts
{ containerClient: FinchClient, orchestratorClient: FinchComposeClient },
```

That single entry registers both clients, adds Finch to the runtime quick pick, and names it in telemetry. The rest:

- Exported both clients from `packages/vscode-container-client/src/index.ts`.
- Added to the `containers.containerClient` and `containers.orchestratorClient` setting enums / `enumItemLabels` (still hand-maintained, as `officialRuntimeRegistrations`' own doc comment notes).
- `getLocalRegistry()` in `netSdkTaskUtils.ts` is **unchanged**; confirmed its `default` branch already returns `undefined` for Finch, which is correct since the .NET SDK's `LocalRegistry` only understands `Docker` and `Podman`. Finch is not routed through #561's tar-archive path either — that's wslc-specific, and matching nerdctl's existing behaviour keeps this PR scoped.

## Tests

- `FinchClient.test.ts` — client ID identity, and version parsing across the real Finch shape, the plain nerdctl shape, a version output with no version fields, and garbage input under both `strict: true` (throws) and `strict: false` (returns `unknown`). The Finch fixture is real captured CLI output, not a hand-written approximation.
- `FinchComposeClient.test.ts` — client ID identity, `composeV2`, and that the emitted command shape matches nerdctl compose.
- `ContainersClientE2E.test.ts` / `ContainerOrchestratorClientE2E.test.ts` now construct `FinchClient` / `FinchComposeClient` for `clientTypeToTest === 'finch'` instead of `new NerdctlClient('finch', ...)`.

### E2E run against a real Finch VM

`CONTAINER_CLIENT_TYPE=finch pnpm testintegration` on Windows/WSL2, Finch v1.17.2:

- **`ContainerOrchestratorClientE2E`: all passing** — Up, Down, Stop, Restart, Logs, ConfigServices against `FinchComposeClient`.
- **`ContainersClientE2E`: 34 passing, 4 failing.** `VersionCommand` and `ClientIdentity` both pass.

The 4 failures are **pre-existing, not regressions**. Confirmed by re-running the suite with the old `new NerdctlClient('finch', ...)` construction, which fails the identical 4:

1. `ReadFileCommand` / `WriteFileCommand` / `WriteFileCommand (streamed)` — one shared upstream nerdctl gap. `finch cp <container>:/path -` fails with `support for writing a tar archive to stdout is not implemented yet`. The *write* direction works (verified via `finch cp` + `exec cat`); both write tests fail only at their tar-based read-back verification.
2. `Containers` "before all" hook — a 10s timeout, not a functional failure. The hook issues 7 sequential finch commands without raising the default timeout (its sibling `after` hook uses 20s); each command pays a ~1.4s VM round-trip, totalling ~11s. This will affect any VM-backed runtime including nerdctl, so I've left it rather than widening this PR's diff.

## Notes for reviewers

**`login`/`logout` work on all platforms.** Finch's `nerdctlCmds` map only adds them when `GOOS == "linux"`, which looks like a Linux-only restriction but isn't: on macOS/Windows Finch registers its own native `newLoginLocalCommand()`/`newLogoutLocalCommand()` in `main_remote.go` instead, so the Linux branch is filling the gap on the one platform that lacks them. Verified `finch login --help` on Windows exposes `--username` and `--password-stdin`, matching what `DockerClientBase.getLoginCommandArgs` emits. No special handling needed. (Side benefit on macOS/Windows: credentials land in the host's Docker config rather than inside the VM.)

**No compose path translation.** Finch rewrites arguments for `container run`, `exec`, and `compose` in `cmd/finch/nerdctl_remote.go`, but its `argHandlerMap` only translates host→VM paths for `container run --mount` and `image build --load`. **There is no `compose` entry**, so `--file`, `--project-directory`, and build contexts referenced inside a compose file are passed into the Lima VM verbatim. `DockerComposeClientBase.withCommonOrchestratorArgs` emits the long form `--file`, which falls through Finch's generic `--`-prefixed branch untouched. Whether an absolute host path resolves inside the VM depends entirely on Lima's mounts. (`--env-file` is fine: Finch reads it on the host and re-expands entries as individual `-e KEY=VALUE` args.)

**The VM does not auto-start.** Finch registers no service, scheduled task, `Run` key, or startup entry, and has no `finch vm` autostart option — the `lima-finch` WSL2 distro starts on demand only. Since the extension has no runtime discovery or VM lifecycle management (`RuntimeManager.getClient()` is a straight settings lookup), a user who selects Finch and reboots will see every container command fail until they manually run `finch vm start`, with no hint as to why. Docker Desktop installs a `Run` key and so doesn't have this problem. Worth a follow-up.

## Out of scope

Compose file path translation and VM lifecycle management (both above), and auto-detection of installed runtimes.

## Verification

- `pnpm build` and `pnpm lint` clean at the repo root.
- All unit tests pass (235 container-client, 154 extension, 130 compose-language-service).
- E2E run against a live Finch VM as detailed above.
- Confirmed the Finch entries appear in both settings enums.


packages/vscode-container-client/src/clients/FinchClient/FinchClient.ts
```diff
@@ -0,0 +1,72 @@
...
+import type { IContainersClient, VersionItem } from '../../contracts/ContainerClient';
+import { NerdctlClient } from '../NerdctlClient/NerdctlClient';
+import { FinchVersionRecordSchema } from './FinchVersionRecord';
+
+/**
+ * A client for the Finch CLI. Finch is a thin wrapper around nerdctl (running in a
+ * Lima VM), forwarding almost every subcommand to nerdctl verbatim, so the command
+ * lines are identical to {@link NerdctlClient}'s.
+ */
+export class FinchClient extends NerdctlClient implements IContainersClient {
```
patverb, line 14: Apparently the login command for finch is only supported for Linux?
https://github.com/runfinch/finch/blob/01fd081001f0d0ea36b181fb42d8766401d1e307/cmd/finch/nerdctl.go#L165
Should we add a better error handling for this otherwise it will just return command not found.

packages/vscode-container-client/src/clients/FinchClient/FinchClient.ts
```diff
@@ -0,0 +1,72 @@
...
+     * The nerdctl version is reported as the client version, for consistency with what
+     * the rest of the extension expects a container client version to mean.
+     */
+    protected override parseVersionCommandOutput(output: string, strict: boolean): Promise<VersionItem> {
+        try {
+            const version = FinchVersionRecordSchema.parse(JSON.parse(output));
+
+            // Prefer the nerdctl version, but fall back to `Client.Version` (the Finch
+            // version) in case `NerdctlClient` is absent, e.g. on older Finch releases
+            const clientVersion = version.Client.NerdctlClient?.Version || version.Client.Version;
```
patverb: checkInstall() and version() will return different values for finch.

`checkInstall` will return Finch version 
`version` will return nerdctlVersion

We dont use versions anywhere it seems but it might be confusing that these methods on a container client could have varied results.

packages/vscode-container-client/src/clients/FinchClient/FinchVersionRecord.ts
```diff
@@ -0,0 +1,34 @@
...
+// `Client.Version` is the Finch version. `Server` is passed through from nerdctl
+// unmodified.
+export const FinchVersionRecordSchema = z.object({
+    Client: z.object({
+        Version: z.optional(z.string()),
+        GitCommit: z.optional(z.string()),
+        GoVersion: z.optional(z.string()),
+        Os: z.optional(z.string()),
+        Arch: z.optional(z.string()),
+        NerdctlClient: z.optional(z.object({
```
patverb: I think this schema is pretty similar to nerdctl minus this field. Could we combine them in some way kind of like we did with the shared schema?

patverb (dismissed):

packages/vscode-container-client/src/clients/FinchClient/FinchClient.ts
bwateratmsft: Good catch — this pushed me to look at how `VersionItem.client` is used across the board, and the answer changed my mind about the whole approach.

Every sibling client reports the version of the CLI you actually invoke: Docker reports docker, Podman reports podman, wslc reports wslc, nerdctl reports nerdctl. And you're right that nothing consumes `version()` today — the only version-adjacent consumer in the extension is `RuntimeInstallStatusProvider`, which calls `checkInstall` and discards the output entirely.

So reporting nerdctl was the odd one out on two counts, and I've flipped it to report the Finch version (0047fe93).

The nice side effect: `Client.Version` *is* the Finch version, so the inherited nerdctl parsing already produces the right answer. The override and `FinchVersionRecord.ts` were both redundant and are now deleted — which also resolves the schema-duplication comment below by removing the duplicate. Net -76/+25.

Verified against Finch v1.17.2, the two now agree:

```
version():      {"client":"v1.17.2","server":"v2.2.1"}
checkInstall(): "finch version v1.17.2"
```

The nerdctl version is no longer surfaced. It's arguably the more *functionally* meaningful number (nerdctl determines which flags actually work), but there's no field for it and `client` semantically means the CLI being invoked, so I've treated it as a VM implementation detail — same as containerd's `runc` component, which we also don't surface.

I kept the version unit tests even though the parsing is now inherited, since Finch's output shape genuinely differs from nerdctl's and it's worth pinning that `client` is `v1.17.2` and not the nested `v2.2.2`. Left a comment on the class documenting the shape so nobody "fixes" it back.

packages/vscode-container-client/src/clients/FinchClient/FinchVersionRecord.ts
bwateratmsft: Agreed — the two schemas were identical apart from the nested `NerdctlClient` field.

It's resolved by deletion rather than dedupe: per the `checkInstall`/`version` thread above, `client` should be the Finch version, which is plain `Client.Version` — exactly what the inherited nerdctl parsing already reads. So the override and this whole file are redundant and are gone as of 0047fe93. No schema left to combine.

If we'd kept the nerdctl-version behavior I'd have factored the shared `Client`/`Server` shape out into `NerdctlVersionRecord.ts` and had Finch compose it, since Finch's nested `NerdctlClient` object has the identical shape to nerdctl's top-level `Client` (including its own `Components` array). Worth remembering if a future Finch-specific field forces the schema back.

packages/vscode-container-client/src/clients/FinchClient/FinchClient.ts
bwateratmsft: I dug into this one and I think we're safe — the conditional is doing the opposite of what it looks like.

That `runtime.GOOS == "linux"` block gates the **nerdctl passthrough map** (`nerdctlCmds`), not the availability of `login`. On non-Linux, Finch registers its own native host-side implementations instead, in [`main_remote.go`](https://github.com/runfinch/finch/blob/01fd081001f0d0ea36b181fb42d8766401d1e307/cmd/finch/main_remote.go):

```go
newLoginLocalCommand(),
newLogoutLocalCommand(),
```

(backed by `login_local.go` / `logout_local.go`). `main_remote.go` is the macOS/Windows entrypoint — the VM-based ones. `main_native.go`, the Linux entrypoint, doesn't register them, which is precisely why `nerdctlCmds` adds `login`/`logout` to the passthrough map for `GOOS == "linux"`: it's filling the gap on the one platform without the local command, not restricting the others.

So login/logout exist on every platform, just via two different routes. Confirmed on my Windows box (Finch v1.17.2):

```
> finch login --help
Log in to a container registry

Usage:
  finch login [flags] [SERVER]

Flags:
      --namespace string   Namespace
  -p, --password string    Password
      --password-stdin     Take the password from stdin
  -u, --username string    Username
```

That matches what `DockerClientBase.getLoginCommandArgs` emits (`login --username <u> --password-stdin <registry>`) exactly, including `--password-stdin`, so no extra error handling is needed and I haven't changed anything here.

Worth noting the routing difference has a real upside on macOS/Windows: because login is handled host-side, credentials land in the host's Docker config rather than inside the VM.

patverb (approved):

PR #577, opened 2026-08-04
Add support for Apple's `container` CLI runtime

### LINKED ISSUE #143: Support for coming macOS container runtime
Yes, it's just in beta, but it's never too early to think about support for the coming macOS 26 [container runtime](https://github.com/apple/container).

VictorPuga:
Relates to #143.

### Summary

Adds `AppleContainerClient`, an `IContainersClient` implementation for Apple's `container` CLI (macOS 26+, Apple Silicon only). Extends `DockerClientBase` but overrides nearly every command builder, since the CLI's surface (bare verbs like `run`/`list`/`stop`, no `--filter`, non-Docker JSON shapes) isn't Docker-compatible enough to inherit as-is. Registered behind `isMac() && isArm64()`, no compose counterpart (none exists), same as `wslc`.

All behavior below was verified against a real CLI 1.2.0 install, not inferred from docs.

### Supported

`checkInstall`, `version`, `info` (synthesized), `run`, `list`, `start`, `stop`, `delete`, `exec`, `logs`, `inspectContainers`, `stats`, `build`, `prune` (container/image/volume/network), `login`/`logout`, `readFile`/`writeFile` (via `exec`+`tar`, since `cp` has no stdin/stdout support), `inspectImages`, `image pull` (pinned to `--arch arm64`), `image list`, full volume and network support (list/inspect/create/remove/prune).

### Notable gotchas handled

- No ID-based image addressing at all — only `name:tag` resolves, so `ListImagesItem.id` is the reference, not the digest.
- Container state is just `running`/`stopped`; mapped to Docker's `running`/`exited` so tree icons and start/stop visibility read correctly.
- `container start` only accepts one ID at a time — multi-select throws instead of silently starting one.
- `logs` has no `--timestamps`/`--since`/`--until`, tails via `-n` not `--tail`.
- `cp` doesn't support `-` for stdin/stdout, so `readFile`/`writeFile` go through `exec ... tar` instead (mirrors the same gap in `WslcClient`).
- `build` and `stats` are top-level verbs (`build`, `stats`), not `image build`/`container stats`; the latter would double the noun prefix and fail to route.
- `prune` needed per-resource handling: the base class's `container container prune` double-noun bug, plus an unconditional `--force` that none of container/image/volume/network prune actually accept.
- `login`/`logout` route through `registry login`/`registry logout`, not a nonexistent top-level `login`/`logout`.
- Network names must be lowercase; `events`/`restart` don't exist and are rejected explicitly.

### Out of scope

- Pre-existing, unrelated gap: "View Logs"/"Build Image" tree commands use a hardcoded Docker-syntax template (`containers.commands.*` in package.json) that bypasses `IContainersClient` entirely, with no per-runtime override. Confirmed broken for this runtime; workaround is a custom `containers.commands.logs` setting. Real fix is a separate, bigger change.
- No Compose equivalent exists for this CLI, so no orchestrator client (same as `wslc`).

### Testing

330 unit tests passing (full package suite), lint clean. E2E suite (`ContainersClientE2E.test.ts`) now wired up for this client: 38 passing / 8 pending against the real `container` CLI on Apple Silicon, and the same suite re-run against Docker (44 passing / 2 pending) to confirm the shared fixture changes didn't regress other runtimes. Manually verified end-to-end on real hardware (pull/run/list/stop/start/inspect/exec/logs/remove/build/volumes/networks) plus the Images/Containers tree views and tooltips in the running extension.


VictorPuga:

extensions/vscode-containers/package.json
```diff
@@ -2594,15 +2594,17 @@
...
                     ],
                     "enumItemLabels": [
                         "Default",
                         "Docker",
                         "Podman",
                         "Nerdctl",
                         "Finch",
-                        "WSLC (Windows only, preview)"
+                        "WSLC (Windows only, preview)",
+                        "Container (macOS only, preview)"
```
bwateratmsft: Does not need to be part of your PR, but I wonder if these can be moved to `package.nls.json`...

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
```diff
@@ -0,0 +1,413 @@
...
+ *   `docker container run`-style. Image verbs do nest under `image`, matching Docker.
+ * - `list` / `image list` accept no `--filter` flag at all (stricter than even `wslc`); all
+ *   filtering in {@link ListContainersCommandOptions} is applied client-side.
+ * - `--format json` is a literal token, not a Go template.
+ * - `image pull` fetches every platform in a multi-arch manifest by default; pinned to
+ *   `--arch arm64` here since this client only ever runs on Apple Silicon.
+ * - No `events`, `restart`, or `info` subcommand exists.
+ * - `--version`/`-v` only accepts the long form; the short form errors.
+ */
+export class AppleContainerClient extends DockerClientBase implements IContainersClient {
```
bwateratmsft, line 132: It's on me to look into this (or Copilot lol), but we should see to what degree we can inherit from the base client. I imagine not a ton since Apple Containers seem to be substantially different from Docker/Podman/Nerdctl/Finch...more like WSLC.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
```diff
@@ -0,0 +1,413 @@
...
+    protected override parseInspectContainersCommandOutput(
+        options: InspectContainersCommandOptions,
+        output: string,
+        strict: boolean,
+    ): Promise<Array<InspectContainersItem>> {
+        return this.parseInspectJson(output, strict, (item) =>
+            normalizeAppleContainerInspectContainerRecord(AppleContainerInspectContainerRecordSchema.parse(item), JSON.stringify(item)));
+    }
+
+    //#endregion
```
patverb, line 591: No Volume or Network section. Many of those commands have missing options that would throw or different outputs that need parsing to be handled like `container volume ls` and `container network ls`.

patverb, line 591: Apple containers do not support login. Should throw an explicit command not supported error.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerInspectContainerRecord.ts
```diff
@@ -0,0 +1,97 @@
...
+        // (ImageTreeItem, ContainerTreeItem, askCopilot) slice this assuming that prefix.
+        imageId: container.configuration.image.descriptor?.digest ?? '',
+        image: parseDockerLikeImageName(container.configuration.image.reference),
+        isolation: undefined,
+        status: undefined,
+        environmentVariables: parseDockerLikeEnvironmentVariables(initProcess?.environment ?? []),
+        networks,
+        ipAddress: networks[0]?.ipAddress,
+        operatingSystem: 'linux',
+        // `configuration.publishedPorts`/`.mounts` shapes haven't been captured against real
```
patverb: Ports and mounts should not be hard coded as empty.

This is wrong. As of 1.2.0 they are there.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerListContainerRecord.ts
```diff
@@ -0,0 +1,77 @@
...
+    return {
+        id: container.id,
+        // The `container` CLI has no name distinct from the container's ID -- `--name` (or
+        // the auto-generated ID) is the same value in both places.
+        name: container.id,
+        labels: container.configuration.labels ?? {},
+        image: parseDockerLikeImageName(container.configuration.image.reference),
+        // `configuration.publishedPorts` shape hasn't been captured against a real `--publish`
+        // run yet; left empty rather than guessing field names. Fill in once verified.
+        ports: [],
```
patverb: Ports should be there. They are present in `container list --format json`

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerInspectImageRecord.ts
```diff
@@ -0,0 +1,97 @@
...
+    const variant = selectPrimaryVariant(image.variants);
+    const config = variant?.config?.config;
+
+    return {
+        // `container` has no ID-based image addressing (see the note in
+        // AppleContainerListImageRecord.ts); mirror that file's `id` choice so this stays a
+        // usable CLI reference rather than an inert digest.
+        id: image.configuration.name ?? image.id,
+        image: parseDockerLikeImageName(image.configuration.name),
+        repoDigests: image.configuration.descriptor?.digest ? [image.configuration.descriptor.digest] : [],
```
patverb: This should be `repository@sha256:....` to match other clients I believe.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerInspectImageRecord.ts
```diff
@@ -0,0 +1,97 @@
...
+        // `image inspect` doesn't distinguish local-only images from ones pulled from a
+        // registry; every inspectable image is on-disk, so this is always true.
+        isLocalImage: true,
+        environmentVariables: parseDockerLikeEnvironmentVariables(config?.Env ?? []),
+        // No ExposedPorts-equivalent field observed in the OCI image config `image inspect`
+        // emits.
+        ports: [],
```
patverb, line 89: Same as other file. these are there under variations[].config.config.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
```diff
@@ -0,0 +1,413 @@
...
+            withArg(options.imageRef),
+            typeof options.command === 'string'
+                ? withVerbatimArg(options.command)
+                : withArg(...(options.command ?? [])),
+        )();
+    }
+
+    // `container run --mount` uses `target=` for the in-container path, not Docker's
+    // `destination=`.
+    protected override getRunContainerMountsArg(mounts: RunContainerCommandOptions['mounts']) {
```
patverb: `--mount destionation=` should work. I think this whole override isnt neeeded and we can use `withDockerMountsArg`

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
```diff
@@ -0,0 +1,413 @@
...
+
+    // `container start` accepts exactly one positional container ID -- confirmed: a second ID
+    // errors with "Unexpected argument '<id>'" and *neither* container ends up started. The
+    // "Start" tree command can multi-select several stopped containers into one
+    // `startContainers({ container: [...] })` call, but a single command invocation here can't
+    // fan out to N separate CLI calls, so reject a multi-container request outright rather than
+    // silently starting only the first (or none).
+    protected override getStartContainersCommandArgs(options: StartContainersCommandOptions): CommandLineArgs {
+        if (options.container.length > 1) {
+            throw new CommandNotSupportedError('container start only supports starting one container at a time.');
```
patverb, line 432: We could just start them one at a time instead of saying that its not supported to start multiple?

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
```diff
@@ -0,0 +1,413 @@
...
+import { AppleContainerListContainerRecordSchema, normalizeAppleContainerListContainerRecord } from './AppleContainerListContainerRecord';
+import { AppleContainerListImageRecordSchema, normalizeAppleContainerListImageRecord } from './AppleContainerListImageRecord';
+
+/**
+ * {@link AppleContainerClient} implements {@link IContainersClient} for Apple's `container`
+ * CLI (macOS 26+, Apple Silicon only -- see https://github.com/apple/container). It extends
+ * {@link DockerClientBase} for its output-parsing helpers, but its command surface is not
+ * Docker-CLI-compatible enough to inherit much else -- most command-building methods are
+ * overridden. All behavior below was verified against real CLI 1.2.0 output; see
+ * `apple-container-poc-plan.md` at the repo root for the raw captures.
```
patverb: You should really remove references to your AI instructions. Its referenced a few times

packages/vscode-container-client/src/test/clients/AppleContainerClient/AppleContainerClient.test.ts
patverb, line 1: Something to consider. Look at what was done for wslc for the `WslcConary.tests.ts`. This asserts the commands that arent implemented fail. That way we know when something new is added and we can support it properly.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
patverb, line 1: You need to fix the `prune` command as now it will call `container container prune` which will fail, it needs to drop the second container. Also '--force doesnt work with `container prune` so we need to drop it for apple container.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
```diff
@@ -0,0 +1,413 @@
...
+        options: ListContainersCommandOptions,
+        output: string,
+        strict: boolean,
+    ): Promise<Array<ListContainersItem>> {
+        return this.parseInspectJson(output, strict, (item) =>
+            normalizeAppleContainerListContainerRecord(AppleContainerListContainerRecordSchema.parse(item)))
+            .then((items) => items.filter((item) => this.matchesListContainersOptions(item, options)));
+    }
+
+    // `imageAncestors`/`volumes`/`networks` filters have no client-side equivalent that can be
```
patverb: This is incorrect. both are in `container list --format json`:

`configration.image.descriptor.digest` -> `imageAncestors`
`configuration.mounts[].types.volume.name` -> `volumes`
`configuraiton.networks[].network` -> networks

These are very much needed for deletion as we use these to see if things are in use.

patverb (changes requested):

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerInspectImageRecord.ts
VictorPuga: Per Claude:

> I verified against redis:7.2-alpine (which has real EXPOSE/VOLUME in its build history) that Apple's image inspect genuinely omits ExposedPorts/Volumes from the OCI config. The current empty-array code is correct as written.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
VictorPuga: Per Claude:

>  the client architecture only lets startContainers() return one command invocation (one args+parse descriptor), not a loop of N real execs; the CommandNotSupportedError guard is the honest option given that constraint.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
VictorPuga: Fixed

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
VictorPuga: Fixed. Sorry about that.

packages/vscode-container-client/src/test/clients/AppleContainerClient/AppleContainerClient.test.ts
VictorPuga: Fixed

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
VictorPuga: Fixed

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerListImageRecord.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    return {
+        // Falls back to the (functionally unusable) digest only for images with no name --
+        // ListImagesItem.id must be a non-empty string, and such images can't be individually
+        // referenced by this CLI at all regardless of what string is put here.
+        id: image.configuration.name ?? image.id,
+        image: parseDockerLikeImageName(image.configuration.name),
+        createdAt: image.configuration.creationDate ?? new Date(0),
+        size: size > 0 ? size : undefined,
+    };
```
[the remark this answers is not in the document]
VictorPuga: Fixed

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerInspectContainerRecord.ts
```diff
@@ -0,0 +1,97 @@
...
+ * {@link InspectContainersItem}.
+ */
+export function normalizeAppleContainerInspectContainerRecord(container: AppleContainerInspectContainerRecord, raw: string): InspectContainersItem {
+    const initProcess = container.configuration.initProcess;
+    const networks: InspectContainersItemNetwork[] = (container.status.networks ?? []).map((network) => ({
+        name: network.network ?? '',
+        gateway: network.ipv4Gateway,
+        ipAddress: network.ipv4Address,
+        macAddress: network.macAddress,
+    }));
```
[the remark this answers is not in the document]
VictorPuga: Fixed

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerInspectContainerRecord.ts
```diff
@@ -0,0 +1,97 @@
...
+        startedDate: z.optional(dateStringWithFallbackSchema),
+        networks: z.optional(z.array(AppleContainerStatusNetworkSchema)),
+    }),
+});
+
+export type AppleContainerInspectContainerRecord = z.infer<typeof AppleContainerInspectContainerRecordSchema>;
+
+function stripDigestPrefix(digest: string | undefined): string {
+    return digest?.replace(/^sha256:/, '') ?? '';
+}
```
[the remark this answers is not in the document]
VictorPuga: Fixed

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
VictorPuga: It exposes `container registry login`

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerInspectContainerRecord.ts
VictorPuga: Fixed

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerListContainerRecord.ts
VictorPuga: Fixed

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerInspectImageRecord.ts
VictorPuga: Fixed

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
VictorPuga: Fixed

TwistedCrafts (approved):
hope this is a fix.

VictorPuga:
Hi, @patverb. The code review comments were addressed.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
bwateratmsft: Agree with @VictorPuga on this one. Single command at a time must be returned. At the extension level, we could loop instead, but I'm not sure we use this much...

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
```diff
@@ -0,0 +1,716 @@
...
+ *   `docker container run`-style. Image verbs do nest under `image`, matching Docker.
+ * - `list` / `image list` accept no `--filter` flag at all (stricter than even `wslc`); all
+ *   filtering in {@link ListContainersCommandOptions} is applied client-side.
+ * - `--format json` is a literal token, not a Go template.
+ * - `image pull` fetches every platform in a multi-arch manifest by default; pinned to
+ *   `--arch arm64` here since this client only ever runs on Apple Silicon.
+ * - No `events`, `restart`, or `info` subcommand exists.
+ * - `--version`/`-v` only accepts the long form; the short form errors.
+ */
+export class AppleContainerClient extends DockerClientBase implements IContainersClient {
```
bwateratmsft, line 132: This client should be wired in to the Container Client E2E tests, and those should be run to ensure that coverage is maximized.

extensions/vscode-containers/package.json
bwateratmsft: Created #582

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
```diff
@@ -0,0 +1,716 @@
...
+    protected override getRemoveContainersCommandArgs(options: RemoveContainersCommandOptions): CommandLineArgs {
+        return composeArgs(
+            withArg('delete'),
+            withFlagArg('--force', options.force),
+            withArg(...options.containers),
+        )();
+    }
+
+    // There is no `restart` subcommand.
+    public override restartContainers(options: RestartContainersCommandOptions): Promise<PromiseCommandResponse<Array<string>>> {
```
bwateratmsft, line 539: 🤖 While you're overriding the inherited container-noun commands here, `statsContainers` needs the same treatment — it's currently broken as inherited. The base builds `container stats` (Docker's `docker container stats`), which for this client becomes `container container stats` and fails with `Error: Plugin 'container-container' not found`. On top of that, `container stats` has no `--all` flag at all — it shows all running containers by default and errors with `Unknown option '--all'`.

It needs a `getStatsContainersCommandArgs` override that returns a bare `stats` with no `--all`. Verified against CLI 1.2.2.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
```diff
@@ -0,0 +1,716 @@
...
+        strict: boolean,
+    ): Promise<Array<InspectImagesItem>> {
+        return this.parseInspectJson(output, strict, (item) =>
+            normalizeAppleContainerInspectImageRecord(AppleContainerInspectImageRecordSchema.parse(item), JSON.stringify(item)));
+    }
+
+    // `image prune` accepts `--all` but not `--force` (confirmed: errors with "Unknown option
+    // '--force'"), unlike the base which always passes `--force`. Real output: a "Reclaimed X
+    // in disk space" summary line, then one `deleted <digest>` line per removed image.
+    protected override getPruneImagesCommandArgs(options: PruneImagesCommandOptions): CommandLineArgs {
```
bwateratmsft, line 330: 🤖 `buildImage` is also broken as inherited. The base builds `image build`, but Apple's `container` has no `image build` subcommand — `build` is a top-level verb, so `container image build ...` errors (`Unknown option '-t'`, and it just routes to the `image` help without building anything).

This needs a `getBuildImageCommandArgs` override that uses a bare `build`, and drops `--iidfile` and `--disable-content-trust` (neither exists on `container build`). Everything else it accepts maps over: `--file`/`--target`/`--tag`/`--label`/`--platform`/`--build-arg`/`--pull` plus the positional context dir. Verified against CLI 1.2.2 — a real `container build --tag x .` succeeds where `container image build` fails.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
```diff
@@ -0,0 +1,716 @@
...
+        )();
+    }
+
+    protected override parseInspectNetworksCommandOutput(
+        options: InspectNetworksCommandOptions,
+        output: string,
+        strict: boolean,
+    ): Promise<Array<InspectNetworksItem>> {
+        return this.parseInspectJson(output, strict, (item) =>
+            normalizeAppleContainerInspectNetworkRecord(AppleContainerListNetworkRecordSchema.parse(item), JSON.stringify(item)));
```
bwateratmsft, line 748: 🤖 `readFile` and `writeFile` are also broken as inherited, and need overrides at the end of the class (a new File Commands region).

The base's `readFile` streams via `cp CONTAINER:PATH -`, and `writeFile` streams via `cp - CONTAINER:DIR`. But Apple's `container cp` supports neither stdin nor stdout `-` — a literal `-` is just treated as a local filename (verified: `container cp c:/etc/hostname -` writes a file named `-` in the cwd and emits nothing; the stdin write form silently doesn't write the piped content).

Suggested fix, mirroring what `WslcClient` does for reads:
- `getReadFileCommandArgs` → `exec <container> tar -cf - -C <dir> <file>`, so the caller still receives the single-entry tarball stream it expects (requires `tar` in the image; Linux containers only).
- `getWriteFileCommandArgs` → for the streamed case (no `inputFile`), `exec -i <container> tar -xf - -C <path>` to extract the incoming tar; when a host `inputFile` is given, fall back to Apple's normal working `cp <inputFile> <container>:<path>`.

Note this diverges from `WslcClient.getWriteFileCommandArgs`, which can still use `cp - CONTAINER:DIR` because wslc's `cp` *does* accept stdin `-`; Apple's does not, so the streamed write has to go through `exec`. All verified against CLI 1.2.2.

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
VictorPuga: Fixed

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
VictorPuga: Fixed

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
VictorPuga: Fixed

packages/vscode-container-client/src/clients/AppleContainerClient/AppleContainerClient.ts
VictorPuga: Done

VictorPuga:
Hi @bwateratmsft and @patverb. Is there anything pending in this PR?

PR #578, opened 2026-08-06
vscode-container-client: parse compacted Docker port ranges

### LINKED ISSUE #572: The Containers view fails to load with "Error: Invalid container JSON" and cannot connect to Docker
**OS:** Win 11 x64
**Build Version:** 20260801.1

**Repro Steps:**
1. Ensure Docker Desktop is running.
2. Install Container Tools extension (latest version) in VS Code.
3. Open Containers view.
4. Observe the Containers section.

**Expect:**
The Containers view loads container resources successfully.

**Actual:**
The Containers view fails to load with "Error: Invalid container JSON". 
Image

**More Info:**
1. The Images, Networks, and other Docker resources are successfully loaded.
2. Configure container runtime with Docker and reload VS Code, this issue still reproduces.

bwateratmsft:
## Summary

- parse consecutive port ranges emitted by `docker container ls`, including exposed and IPv4/IPv6 published ranges
- expand compacted ranges into individual port bindings during container normalization
- add regression coverage for the Azurite-style `10000-10002` range and malformed ranges

## Validation

- 239 unit tests passed
- `pnpm --filter @microsoft/vscode-container-client run lint`
- `pnpm --filter @microsoft/vscode-container-client run build`

Fixes #572

### DIFF
--- a/packages/vscode-container-client/src/clients/DockerClientBase/parseDockerRawPortString.ts
+++ b/packages/vscode-container-client/src/clients/DockerClientBase/parseDockerRawPortString.ts
@@ -36,6 +36,7 @@ export function parseExposedPortKey(key: string): { containerPort: number; proto
 }
 
 const shortFormRegex = /^(?<containerPort>\d+)\/(?<protocol>tcp|udp)$/i;
+const shortRangeFormRegex = /^(?<containerPortStart>\d+)-(?<containerPortEnd>\d+)\/(?<protocol>tcp|udp)$/i;
 
 // Supports:
 // - hostPort->containerPort[/protocol]
@@ -46,6 +47,7 @@ const shortFormRegex = /^(?<containerPort>\d+)\/(?<protocol>tcp|udp)$/i;
 //   `:` before the host port, so embedded IPv6 colons are preserved; brackets
 //   (if any) are stripped by normalizeIpAddress.
 const longFormRegex = /^(?:(?<host>\[[^\]]*\]|[^\s]*?):)?(?<hostPort>\d+)\s*->\s*(?<containerPort>\d+)(?:\/(?<protocol>tcp|udp))?$/i;
+const longRangeFormRegex = /^(?:(?<host>\[[^\]]*\]|[^\s]*?):)?(?<hostPortStart>\d+)-(?<hostPortEnd>\d+)\s*->\s*(?<containerPortStart>\d+)-(?<containerPortEnd>\d+)(?:\/(?<protocol>tcp|udp))?$/i;
 
 /**
  * Attempt to parse a Docker-like raw port binding string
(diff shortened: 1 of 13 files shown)


packages/vscode-container-client/src/clients/DockerClientBase/parseDockerRawPortString.ts
```diff
@@ -81,3 +84,70 @@ export function parseDockerRawPortString(portString: string): PortBinding | unde
...
+            containerPort: containerPortStart + offset,
+            protocol,
+        }),
+    );
+}
+
+function isValidPortRange(start: number, end: number): boolean {
+    return Number.isSafeInteger(start)
+        && Number.isSafeInteger(end)
+        && start >= 0
```
patverb, line 121: Port 0 isnt valid

packages/vscode-container-client/src/clients/DockerClientBase/parseDockerRawPortString.ts
```diff
@@ -36,6 +36,8 @@ export function parseExposedPortKey(key: string): { containerPort: number; proto
 }
 
 const shortFormRegex = /^(?<containerPort>\d+)\/(?<protocol>tcp|udp)$/i;
+const shortRangeFormRegex = /^(?<containerPortStart>\d+)-(?<containerPortEnd>\d+)\/(?<protocol>tcp|udp)$/i;
```
patverb: Consider: Could you combine shortRange and shortForm by making the range part optional? This way you could have a single function that could handle both and you dont need to call it out specifically? (Same for longForm and longRange)

If not does anyone still use the old ParseDockerRawPortStrings? We could just get rid of that too. I think its just called by tests.

patverb: Apparently docker can have protocol `sctp` in addition to `udp` and `tcp`.

packages/vscode-container-client/src/clients/DockerClientBase/parseDockerRawPortString.ts
```diff
@@ -81,3 +84,70 @@ export function parseDockerRawPortString(portString: string): PortBinding | unde
         protocol,
     };
 }
+
+/**
+ * Parse a Docker-style port string into one or more bindings. Docker compacts
+ * consecutive ports into ranges in `ps` output, such as `10000-10002/tcp` or
+ * `0.0.0.0:10000-10002->10000-10002/tcp`.
+ */
+export function parseDockerRawPortStringList(portString: string): PortBinding[] | undefined {
```
patverb: Nitpick: The name feels more like we have a list of string ports that we are parsing. Something like `expandDockerRawPortString` sounds better

packages/vscode-container-client/src/clients/DockerClientBase/parseDockerRawPortString.ts
```diff
@@ -36,6 +36,8 @@ export function parseExposedPortKey(key: string): { containerPort: number; proto
 }
 
 const shortFormRegex = /^(?<containerPort>\d+)\/(?<protocol>tcp|udp)$/i;
+const shortRangeFormRegex = /^(?<containerPortStart>\d+)-(?<containerPortEnd>\d+)\/(?<protocol>tcp|udp)$/i;
+const maxPort = 65535;
```
patverb: Nit: This probably should be near `isValidPortRange` and not between the port parsing regex.

packages/vscode-container-client/src/test/clients/DockerClientBase/parseDockerRawPortString.test.ts
```diff
@@ -67,4 +67,28 @@ describe('(unit) parseDockerRawPortString', () => {
             expect(parseDockerRawPortString(input)).to.be.undefined;
         });
     });
+
```
patverb, line 74: Consider: A few more tests in here. Situation like:
`0.0.0.0:32768-32770->8000-8002/tcp` - different host/container incase they are wrong
`0.0.0.0:100-102-> 100-102/tcp` - ip with port range
`100-101->100-101/tcp` - host and container with ip
`:::10000-10001->10000-10001/udp` - unbracketed ipv6
`10000-10000/tcp` - single port range

packages/vscode-container-client/src/test/clients/DockerClientBase/SharedListContainerRecord.test.ts
```diff
@@ -81,6 +81,26 @@ describe('(unit) normalizeListContainerRecord', () => {
         expect(result.name).to.equal('first');
     });
 
+    it('Should normalize compacted Docker port ranges in strict mode', () => {
+        const parsed = SharedListContainerRecordSchema.parse({
+            ID: 'abc123',
+            Names: 'azurite',
+            Image: 'mcr.microsoft.com/azure-storage/azurite',
+            Ports: '0.0.0.0:10000-10002->10000-10002/tcp, [::]:10000-10002->10000-10002/tcp',
```
patverb: Consider: adding a single port to the list of range ports to make sure it handles both at the same time

packages/vscode-container-client/src/clients/DockerClientBase/parseDockerRawPortString.ts
bwateratmsft: It's not, but I could see some runtime representing an intentionally-invalid port as 0. IMO, better to go ahead and parse it; the data is coming from one of the runtimes, so in theory it should be considered valid and trusted.

packages/vscode-container-client/src/clients/DockerClientBase/parseDockerRawPortString.ts
```diff
@@ -7,15 +7,17 @@ import type { PortBinding } from '../../contracts/ContainerClient';
...
  * (case-insensitive) or undefined for anything else (including undefined input).
  */
-export function normalizeProtocol(protocol: string | undefined): 'tcp' | 'udp' | undefined {
+export function normalizeProtocol(protocol: string | undefined): PortBinding['protocol'] {
     switch (protocol?.toLowerCase()) {
         case 'tcp':
             return 'tcp';
         case 'udp':
             return 'udp';
+        case 'sctp':
```
patverb, line 19: We need to update vscode-containers `package.json` to allow for users to specify sctp. Also `package.nls.json` string "Specify protol for the bind ('tcp | udp')"

patverb (approved):

PR #588, opened 2026-08-20
Add configurable Compose task terminal closing

### LINKED ISSUE #350: Add presentation options to Docker extension's compose commands
Type: Feature Request

When running Docker Compose commands via the Docker extension UI (e.g., right-clicking a container and selecting "Compose Stop"), the terminal shows:

"* Terminal will be reused by tasks, press any key to close it."

This requires manual intervention to close the terminal after each command finishes.

Custom VS Code tasks support a `presentation` property to auto-close terminals:
```json
"presentation": {
  "close": true
}

### PR DESCRIPTION
**Pull request by LE0-Lin**
## Summary

- add an advanced `containers.closeComposeTaskTerminal` setting, disabled by default
- apply the setting to Compose commands and Compose group lifecycle actions
- keep long-running Compose Logs usable: task presentation `close` only takes effect after the task ends and does not interrupt streaming output
- map task focus and close presentation options directly so omitted values stay unset and explicit `false` values are preserved

## Context

This follows up on #363, which was closed due to inactivity, and incorporates the maintainer feedback from that review: make terminal closing opt-in and keep the default behavior unchanged.

Closes #350

## Testing

- `pnpm --filter vscode-containers lint`
- `pnpm --filter vscode-containers build`
- `pnpm --filter vscode-containers test` (161 passing on VS Code 1.136.0)

The task presentation regression tests cover omitted options, explicit `false` values, and enabling terminal closing without changing focus.

## AI disclosure

Codex (GPT-5) assisted with implementation and validation. The contributor reviewed the changes and test results before submission.

### DIFF
--- a/extensions/vscode-containers/package.nls.json
+++ b/extensions/vscode-containers/package.nls.json
@@ -206,6 +206,7 @@
     "vscode-containers.config.docker.languageserver.formatter.ignoreMultilineInstructions": "Controls whether the Dockerfile formatter should ignore instructions that span multiple lines when formatting",
     "vscode-containers.config.containers.composeBuild": "Set to true to include --build option when compose command is invoked",
     "vscode-containers.config.containers.composeDetached": "Set to true to include --d (detached) option when compose command is invoked",
+    "vscode-containers.config.containers.closeComposeTaskTerminal": "Automatically close the terminal after a Compose command run by Container Tools completes.",
     "vscode-containers.config.containers.showRemoteWorkspaceWarning": "Set to true to prompt to switch from \"UI\" extension mode to \"Workspace\" extension mode if an operation is not supported in UI mode.",
     "vscode-containers.config.containers.scaffolding.templatePath": "The path to use for scaffolding templates.",
     "vscode-containers.config.containers.containerCommand": "Command to use for container actions (e.g. `docker` command). If the executable path contains whitespace, it needs to be quoted appropriately. If unset, the extension will attempt to auto-detect the command to use.",
--- a/extensions/vscode-containers/package.nls.json
+++ b/extensions/vscode-containers/package.nls.json
@@ -206,7 +206,7 @@
     "vscode-containers.config.docker.languageserver.formatter.ignoreMultilineInstructions": "Controls whether the Dockerfile formatter should ignore instructions that span multiple lines when formatting",
     "vscode-containers.config.containers.composeBuild": "Set to true to include --build option when compose command is invoked",
     "vscode-containers.config.containers.composeDetached": "Set to true to include --d (detached) option when compose command is invoked",
-    "vscode-containers.config.containers.closeComposeTaskTerminal": "Automatically close the terminal after a Compose command run by Container Tools completes.",
+    "vscode-containers.config.containers.closeComposeTaskTerminal": "Automatically close the terminal after a Compose command run by Container Tools is finished.",
     "vscode-containers.config.containers.showRemoteWorkspaceWarning": "Set to true to prompt to switch from \"UI\" extension mode to \"Workspace\" extension mode if an operation is not supported in UI mode.",
     "vscode-containers.config.containers.scaffolding.templatePath": "The path to use for scaffolding templates.",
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by LE0-Lin** [comment]

File: extensions/vscode-containers/src/runtimes/runners/TaskCommandRunnerFactory.ts [hunk 1]
```diff
@@ -57,11 +58,10 @@ async function executeAsTask(options: TaskCommandRunnerOptions, command: string,
...
 
-    if (options.focus) {
-        task.presentationOptions = {
-            focus: true,
-        };
-    }
+    task.presentationOptions = {
+        focus: options.focus,
+        close: options.close,
+    };
```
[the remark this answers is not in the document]
bwateratmsft: We should test to be sure but I don't think I agree with this feedback...it would be odd if VSCode was behaving wrong from us unconditionally setting this.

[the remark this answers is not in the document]
LE0-Lin: Thanks — I verified this against the VS Code task host and added regression coverage for all three cases: omitted options remain unset, explicit false values are preserved, and enabling close does not change focus. The full extension test suite passes with 161 tests.

extensions/vscode-containers/package.nls.json
```diff
@@ -206,6 +206,7 @@
     "vscode-containers.config.docker.languageserver.formatter.ignoreMultilineInstructions": "Controls whether the Dockerfile formatter should ignore instructions that span multiple lines when formatting",
     "vscode-containers.config.containers.composeBuild": "Set to true to include --build option when compose command is invoked",
     "vscode-containers.config.containers.composeDetached": "Set to true to include --d (detached) option when compose command is invoked",
+    "vscode-containers.config.containers.closeComposeTaskTerminal": "Automatically close the terminal after a Compose command run by Container Tools completes.",
```
bwateratmsft:
```suggestion
    "vscode-containers.config.containers.closeComposeTaskTerminal": "Automatically close the terminal after a Compose command run by Container Tools is finished.",
```

Reads a little easier IMO but I don't feel strongly

extensions/vscode-containers/src/commands/containers/composeGroup.ts
```diff
@@ -30,19 +31,19 @@ export async function composeGroupLogs(context: IActionContext, node: ContainerG
     }, node, { follow: true, tail: 1000 });
 }
 export async function composeGroupStart(context: IActionContext, node: ContainerGroupTreeItem): Promise<void> {
-    return composeGroup(context, (client, options) => client.start(options), node);
+    return composeGroup(context, (client, options) => client.start(options), node, undefined, shouldCloseComposeTaskTerminal());
```
bwateratmsft: I think just look up `shouldCloseComposeTaskTerminal()` within `composeGroup()`; so that you don't have to add an argument and change all these lines to pass it in

bwateratmsft: Additionally, since you would be doing the lookup in exactly one place, you may as well inline the body of `shouldCloseComposeTaskTerminal()`, where it is used on line 89

LE0-Lin: Thanks — updated this in 044aaa6d. composeGroup now reads closeComposeTaskTerminal directly when constructing TaskCommandRunnerFactory, and the extra argument plus helper function have been removed. I also hardened the task presentation regression test to avoid a fast-process event race. Lint, build, and all 161 extension tests pass.

extensions/vscode-containers/package.nls.json
LE0-Lin: Thanks — updated the wording in 044aaa6d to the suggested text. The extension lint, build, and full test suite all pass (161 tests).

LE0-Lin:
Hi @bwateratmsft — I’ve rebased this PR onto the current `main`; the patch itself is unchanged. I reran the relevant checks locally:

- `pnpm --filter vscode-containers lint`
- `pnpm --filter vscode-containers build`
- `pnpm --filter vscode-containers test` (195 passing on VS Code 1.138.0)

The new fork workflow is awaiting maintainer approval. When you have a chance, could you approve it and take another look at the PR? Thanks!

PR #608, opened 2026-09-10
Add Integrated Browser action for running containers

### LINKED ISSUE #600: [FEAT] Add command to use Integrated Browser for "Open In Browser" from running container
### Problem
In VS Code Container Tools, Open In Browser opens containers in External Browser.  Following the recent feature addition of "Integrated Browser" to VS Code, I am requesting an option to use that as the default browser for Open In Browser.

### Why this matters
To minimize context switching, we can use the Integrated Browser in side by side views.

### Requested feature
Add a setting such as:

```json
{
    "containers.useIntegratedBrowser" = false
}
```

Suggested behavior:
1. If true, use VS Code Integrated Browser
2. If false, use Extension default (via ```vscode.env.openExternal```)
3. If not set, use Extension default (via ```vscode.env.openExternal```)

### Environment
- Extension: ms-azuretools.vscode-containers
- OS: MacOS

[Disscussion #599](https://github.com/microsoft/vscode-containers/discussions/599)

Image

bwateratmsft:
🤖
## Summary

Fixes #600 using two explicit browser actions, as discussed in the issue, rather than a preference setting.

- Rename the existing action to **Open in External Browser**, preserving its command ID and behavior.
- Add **Open in Integrated Browser** to the running-container context menu and Command Palette, sharing container and port selection with the external action.
- Resolve browser URLs with `asExternalUri` for remote port forwarding and preserve query parameters when opening the Integrated Browser.
- Raise Container Tools' minimum VS Code version and shared `@types/vscode` to **1.109.0**. Adapt an authentication-error localization argument to use the provider's constructor name for compatibility with the newer types.
- Update the README and add browser-routing regression coverage.

## Validation

- Workspace builds, Container Tools type checking, and targeted lint passed.
- 33 browser regression tests passed on VS Code 1.109.0 and current stable (1.137.0).
- A live HTTP container loaded successfully in the actual Integrated Browser on VS Code 1.109.0.
- Live remote-host and macOS behavior were not exercised.

Lovely example view, with Copilot interacting too:

image


fiveisprime (approved):
:shipit:

alexweininger (approved):

PR #617, opened 2026-09-21
Fix WSLC Docker-style container list parsing

### LINKED ISSUE #616: WSLC: Running Containers View Broken
Image

Output when "container tools" is selected shows

```
2026-09-21 20:56:47.641 [error] Failed to set up event listener: wslc does not support the events command.
```

When I hover on the error message I see more info but can't copy it out, so a screenshot will have to do.

Image

```
> wslc --version     
wslc 2.9.12.0
```

I'm running this version of the extension:-

```
Installation
Identifier
ms-azuretools.vscode-containers
Version
2.5.1
```

bwateratmsft:
🤖
## Summary
- Accept Docker-style container records emitted by WSLC 2.9.12.0 alongside the legacy native format.
- Reuse shared Docker normalization for creation dates, published ports, labels, networks, states, and statuses while preserving strict-mode behavior and array/JSONL support.
- Add synthetic regression fixtures matching both reported output shapes, covering EDT/BST timestamps, IPv4/IPv6 ports, JSON-valued labels, running/exited containers, and malformed records.

Unsupported `events` handling is unchanged.

Fixes #616

## Validation
- Focused WSLC and shared-container unit tests: 95 passing.
- WSLC suite rerun after fixture sanitization: 86 passing.
- Container-client and processutils ESM/CJS builds passed.
- Targeted ESLint and `git diff --check` passed.


packages/vscode-container-client/src/test/clients/WslcClient/WslcClient.test.ts
```diff
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { type CommandLineArgs, NoShell } from '@microsoft/vscode-processutils';
+import { rejects } from 'assert';
```
patverb: No other test in `packages/vscode-container-client/src/test` imports `assert` — the package is consistently chai (` expect(...).to.throw()` , e.g.  DockerClientBase.test.ts:83 ,  SharedListContainerRecord.test.ts:128 ). `chai-as-promised` is already hoisted in  pnpm-workspace.yaml , so this can be  await expect(response.parse(output, true)).to.be.rejectedWith(/…/) .

Separately, bare `rejects(...)` with no matcher passes on any rejection, including an unrelated  TypeError  — worth asserting the message either way, as the ports test already does.

packages/vscode-container-client/src/clients/WslcClient/WslcClient.ts
```diff
@@ -554,4 +554,3 @@ export class WslcClient extends DockerClientBase {
 
     //#endregion
 }
-
```
patverb, line 557: nit: removed EOF newline

patverb (dismissed):

packages/vscode-container-client/src/clients/WslcClient/WslcClient.ts
bwateratmsft: I checked, it appears there were two trailing newlines before this.

packages/vscode-container-client/src/test/clients/WslcClient/WslcClient.test.ts
bwateratmsft: 🤖 Replaced Node's `assert.rejects` with `chai-as-promised`, registered in the WSLC suite. The malformed-record test now requires a Zod `$ZodError` with `invalid_union`, so an unrelated `TypeError` cannot satisfy it; the ports test checks the exact error message. Both assertions invoke parsing through a promise boundary because this parser can throw synchronously despite its Promise return type. Pushed in bb8d8bdf.

patverb (approved):

PR #621, opened 2026-09-23
Add containers-canvas: a Copilot plugin for local Docker

patverb:
Adds **Containers for Copilot** — a canvas plugin that puts your local Docker containers and images in a live panel, so Copilot can answer container questions by showing you the thing instead of pasting command output.

Ask "what's running?" and the panel opens. Ask "why did `web` stop?" and it opens on that container's logs.

image

## What you can do

| View | What it gives you |
| --- | --- |
| **List** | Containers and images, sortable and filterable, with start/stop/restart/pause/remove |
| **Logs** | Live `--follow` streaming with filtering, adjustable tail, copy and save |
| **Stats** | CPU, memory and network as sparklines, sampled about once a second |
| **Files** | Browse a container's filesystem, read files, open one in the Copilot editor |
| **Terminal** | An interactive shell with a real PTY *(optional install)* |
| **Commands** | Run one-off commands, with an auditable record of everything that ran |
| **Layers** | Which layer made an image large, in build order |
| **Dockerfile** | The recorded source repo, plus a labelled reconstruction |
| **Run image** | Start a container from an image, with validated options |

Images can also be pulled and tagged from the panel.

image

## It stays live

The panel tracks the daemon, not a snapshot. Start a container in another terminal, run `docker compose up`, let something crash — it shows up in one to three seconds without being asked.

A shared `docker events` stream does the work; a 10-second poll is only a safety net for what a stream cannot report (dropped connection, daemon restart, waking from sleep). On an idle machine it does no work at all, and actions the panel itself causes are ignored, so opening a folder doesn't make it reload itself.

## Copilot drives it too

The plugin ships a skill that routes container questions to the right view, so the agent lands on logs for "why did it stop?" and on layers for "why is this image 1.2 GB?". Copilot can also run the actions itself — and everything it runs shows up in the panel's Commands history, so you can see what happened rather than trusting a summary.

Destructive actions (remove, prune) ask first, whether they come from you or from Copilot.

## Trying it

```
copilot plugin marketplace add microsoft/vscode-containers
copilot plugin install containers@vscode-containers
```

Then start a new Copilot session — plugins are discovered at session start.

Until this lands on `main`, point the marketplace at a local clone instead; a local path is read live from your working tree:

```
copilot plugin marketplace add <path to your clone>
```

Requires Docker on `PATH`. The panel itself is a GitHub Copilot app feature; on other clients (Copilot CLI, VS Code) the plugin installs and the skill answers the same questions with `docker`.

## Notes for review

This is a Copilot plugin rather than a VS Code extension, so it doesn't change the `vscode-containers` extension or any existing package — it just lives in the same repo, as the same product area.

Being a plugin does mean it breaks one repo convention: `bundle/` and `NOTICE.html` are **committed**, because plugins install straight from a git ref with no build step, so the committed bytes are what users run. That's what the `.gitignore` / `.gitattributes` / `postinstall` changes are for.

Verified with the full CI sequence locally — repo-wide lint, build, package and test, 716 tests — plus a browser harness that drives the real panel over CDP (9/9), and a Linux container build to check cross-platform behaviour. Untested on macOS and Podman.

`extensions/containers-canvas/docs/constraints.md` explains each non-obvious constraint with the incident behind it, if you hit something in the diff that looks arbitrary.


extensions/containers-canvas/extension.mjs
```diff
@@ -0,0 +1,201 @@
...
+                    containers: kind === "images" ? undefined : state.containers.slice(0, limit),
+                    images: kind === "containers" ? undefined : state.images.slice(0, limit),
+                };
+            },
+        },
+        ...buildCanvasActions({
+            describe: describeAgentActions,
+            call: (instanceId, name, input) => instanceFor(instanceId).callProcedure(name, input),
+            instanceFor,
+        }),
```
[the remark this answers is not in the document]
patverb: Fixed in 1e4dd484. `containerOp` and `imageOp` now refuse `remove`/`forceRemove` unless the call sets `acknowledgeDestructive`, mirroring the existing `acknowledgeHostAccess` on `execInContainer`.

Being straight about what that is and is not: it is **not** a boundary against a caller that means it, since the agent can set the flag itself. What it stops is the accidental case — every other verb in that enum is reversible, removal is not, so removal has to be asked for by name rather than arrived at through a mistyped or guessed `op`. The panel supplies the flag only after its confirmation dialog, and `SKILL.md` now tells the agent to ask the user first and only then repeat the call with it.

Covered by four tests in `security.test.mjs`. `verify:panel` still passes 12/12, which is what proves the panel's own path supplies the flag.

extensions/containers-canvas/src/appRouter.mjs
```diff
@@ -0,0 +1,652 @@
...
+        refresh: trpc.publicProcedure
+            .input(z.object({ force: z.boolean().optional() }).optional())
+            .mutation(async ({ input }) => cache.refresh({ force: input?.force ?? true })),
+
+        containerOp: trpc.publicProcedure
+            .input(z.object({ op: containerOps, id: targetId }))
+            .mutation(async ({ input }) => {
+                const result = await containerOp(input.op, input.id);
+                await cache.refresh({ force: true });
+                return result;
```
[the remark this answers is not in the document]
patverb: Fixed in 1e4dd484, same change as the `extension.mjs` thread: both procedures now call `assertDestructiveAcknowledged`, which refuses `remove` and `forceRemove` without an explicit `acknowledgeDestructive`.

The honest limit is the same one `acknowledgeHostAccess` has — an agent that intends to remove something can set the flag. It is a guard against the unintended call, not against intent.

extensions/containers-canvas/scripts/verifyBundle.mjs
```diff
@@ -0,0 +1,110 @@
...
+const { digest, fileCount } = await computeInputDigest(packageRoot);
+
+if (recorded.digest !== digest) {
+    console.error(
+        "\n[verify] the committed bundle was built from different source than is here now:\n" +
+        `  recorded ${String(recorded.digest).slice(0, 16)} over ${recorded.fileCount} files\n` +
+        `  current  ${digest.slice(0, 16)} over ${fileCount} files\n\n${REBUILD_HINT}`,
+    );
+    process.exit(1);
+}
```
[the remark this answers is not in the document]
patverb: Good catch, and this was a regression I introduced. Reproduced before fixing: deleting `bundle/webview/main.js` left the input digest valid and `package` exited 0.

Fixed in 7b3c0dd0. The build now records an inventory of every emitted file with its hash in `bundle/build-inputs.json`, and the check compares it — reporting `missing`, `changed since the build`, and `present but not recorded` separately. On top of that, a git pass treats a **worktree** deletion or an untracked file under `bundle/` as fatal.

What it deliberately still tolerates is a *modified* bundle file, because that is the platform-dependent byte difference the digest exists to work around, and a *staged* deletion, because renaming content-hashed chunks is what a correct rebuild commit looks like. Getting that distinction wrong was a false positive I hit while testing this: `driftIn` had been trimming the porcelain status columns, which collapses "staged for deletion" and "missing from disk" into the same string.

Verified all three ways: deleted file, hand-edited file, and uncommitted chunk each fail; a correctly staged rebuild passes.

extensions/containers-canvas/src/runtime.mjs
```diff
@@ -0,0 +1,1693 @@
...
+    );
+    if (probe.ok) {
+        const resolved = probe.output.trim().split("\n").pop()?.trim();
+        if (resolved && resolved.startsWith("/") && resolved !== target) {
+            source = resolved;
+            resolvedFrom = target;
+        }
+    }
+
+    const base = path.basename(target) || "file";
```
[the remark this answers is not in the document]
patverb: Confirmed and fixed in 7b3c0dd0. Reproduced first:

```
"/.."     basename= ".."  hostPath= C:\dest        <- the extraction root
"/a/.."   basename= ".."  hostPath= C:\dest
"/etc/passwd" basename= "passwd" hostPath= C:\dest\cid\passwd
```

So the directory branch's recursive `rm` would have taken the shared extraction directory with it.

Extracted `extractionBasename`, which refuses `.`, `..` and any separator, and the caller now asserts the result stays under the container's folder. It also splits with **POSIX** rules: `target` is a path inside a Linux container, where a backslash is an ordinary filename character, so `path.basename` on Windows would both truncate the name and, once joined again, let `a\..\..\evil` climb.

Covered by two tests in `runtime.test.mjs`.

extensions/containers-canvas/src/runtime.mjs
```diff
@@ -0,0 +1,1693 @@
...
+
+/** Reasons this container is not meaningfully isolated from the host. */
+function hostEscapeRisks(detail) {
+    const host = detail?.HostConfig ?? {};
+    const risks = [];
+    if (host.Privileged) risks.push("it runs privileged");
+    if ((host.Binds ?? []).some((b) => HOST_ESCAPE_BINDS.test(String(b)))) {
+        risks.push("it mounts the container runtime socket");
+    }
+    if ((host.CapAdd ?? []).some((c) => /SYS_ADMIN|SYS_PTRACE|SYS_MODULE/i.test(String(c)))) {
```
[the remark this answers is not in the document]
patverb: Correct — fixed in 7b3c0dd0. `--cap-add ALL` grants SYS_ADMIN along with everything else, so the broadest possible grant was the one a list of individual names missed.

Capabilities are now normalised (trimmed, upper-cased, optional `CAP_` prefix stripped) and compared against a set containing `ALL`, `SYS_ADMIN`, `SYS_PTRACE` and `SYS_MODULE`. The reported message also now lists only the offending capabilities rather than the whole `CapAdd` array.

Tests cover `ALL`, `CAP_ALL`, `cap_all`, `ALL` alongside a benign capability, and a negative case asserting `NET_BIND_SERVICE`/`CHOWN` stay unflagged.

extensions/containers-canvas/src/runtime.mjs
```diff
@@ -0,0 +1,1693 @@
...
+    /^[a-zA-Z]:[\\/]+(windows|program files|program files \(x86\)|programdata)([\\/]|$)/i,
+    /^[a-zA-Z]:[\\/]+users[\\/]*$/i, // all user profiles; subdirectories are fine
+    /^[\\/](etc|usr|bin|sbin|boot|dev|proc|sys|lib|lib64)([\\/]|$)/i,
+    /docker[._-]?(sock|engine)/i, // docker socket / named pipe == host root
+    /^\\\\[.?]\\pipe\\/i, // Windows named pipes
+];
+
+function assertSafeMount(source) {
+    const value = String(source).trim();
+    if (!value) throw new Error("A volume mount needs a host path.");
```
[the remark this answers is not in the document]
patverb: Confirmed and fixed in 7b3c0dd0. Probing the exported guard found your case and one more:

```
blocked  "/etc"
ALLOWED  "/tmp/../etc"        <- yours
ALLOWED  "//etc/passwd"       <- same rule, doubled separator
blocked  "/etc/./passwd"
```

Two changes. A `..` segment is now refused outright rather than resolved — canonicalising would mean resolving for the daemon's platform rather than this process's, and a bind mount has no need to climb. And the deny-list is matched against a separator-collapsed form as well as the raw value, so `//etc` cannot slip past a rule anchored on one leading separator. The collapse deliberately preserves a leading `\\`, because the named-pipe rule depends on it.

Three tests added, including one asserting an ordinary UNC share still works while `\\.\pipe\docker_engine` is still refused.

extensions/containers-canvas/src/runtime.mjs
```diff
@@ -0,0 +1,1693 @@
...
+ * @param {string} [spec.restart]
+ * @param {string} [spec.memory]        "512m"
+ * @param {string} [spec.cpus]          "1.5"
+ * @param {boolean} [spec.detach=true]
+ * @param {boolean} [spec.removeOnExit=false]
+ * @param {boolean} [spec.publishAll=false]
+ * @param {boolean} [spec.pull=false]
+ */
+export async function runImage(spec = {}) {
+    const image = String(spec.image ?? "").trim();
```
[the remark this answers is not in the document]
patverb: Agreed, and this was the most serious of the set. Fixed in 7b3c0dd0.

`docker run` parses options until its first positional, so an image of `--privileged` was read as a flag and `command[0]` became the image — handing back a privileged container while every mount, port and capability check above had passed. `runImage` now calls `assertImageRef`, the same guard `tagImage`, `pullImage` and `pushImage` already use, and whose own existing test says it "refuses anything that could become another argument". The contract was there; this one call site just never used it.

extensions/containers-canvas/server.mjs
```diff
@@ -0,0 +1,553 @@
...
+        let sessionId = null;
+        try {
+            const started = await execs.start({
+                containerId: target.id,
+                cols: Number(new URL(req.url, "http://127.0.0.1").searchParams.get("cols")) || 80,
+                rows: Number(new URL(req.url, "http://127.0.0.1").searchParams.get("rows")) || 24,
+                onData: (chunk) => send({ type: "data", data: chunk }),
+                onExit: (code) => { send({ type: "exit", code }); socket.close(); },
+            });
+            sessionId = started.id;
```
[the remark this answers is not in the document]
patverb: Confirmed and fixed in 7b3c0dd0. `execs.start` probes the container for a usable shell and then spawns a PTY, which is easily long enough for a panel to navigate away mid-flight — and with the handlers registered afterwards that close fired while nothing was listening.

Ownership is now claimed before the await: `close`/`error` are bound first and set a `socketClosed` flag, and when `start` resolves the session is killed immediately if the socket has already gone. The message handler moved above the teardown so there is a single `done` that both paths share.

extensions/containers-canvas/src/webview/LogView.jsx
```diff
@@ -0,0 +1,332 @@
...
+        setText("");
+        let carry = "";
+        const source = new EventSource(panelHref("./logs", { id: item.id, tail }));
+
+        source.addEventListener("message", (event) => {
+            let frame;
+            try { frame = JSON.parse(event.data); } catch { return; }
+            if (frame.type === "error") { setStreamError(frame.message); return; }
+            if (frame.type === "end") { setFollowing(false); return; }
+            if (frame.type !== "log") return;
```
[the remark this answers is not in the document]
patverb: Confirmed and fixed in 7b3c0dd0. A container that exits without a trailing newline left its last line sitting in `carry` — which is exactly the line someone following logs is waiting for.

The append logic is now a single `append()` used by both paths, with a `flush()` that emits whatever is held back. It runs on the `end` frame and also on the `error` path, since a stream that drops mid-follow has the same partial line pending.

extensions/containers-canvas/src/webview/TerminalView.jsx
```diff
@@ -0,0 +1,215 @@
...
+    const connect = useCallback(() => {
+        const term = termRef.current;
+        if (!term) return;
+        setError(null);
+        setStatus("connecting");
+
+        const fit = fitRef.current;
+        try { fit?.fit(); } catch { /* not laid out yet */ }
+
+        const url = panelUrl("./exec", { id: item.id, cols: term.cols, rows: term.rows });
```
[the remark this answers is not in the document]
patverb: Fixed in 7b3c0dd0. `connect()` now closes the existing socket before replacing `socketRef.current`, guarded on `readyState <= OPEN` so it covers the connecting case you describe — which is the one that matters, since the server may already be starting a PTY that nothing on this side could then reach or shut down.

Chose closing over disabling the button: reconnect is most useful precisely when a connection is wedged part-way, so disabling it during `connecting` would take the control away at the moment it is wanted. The server-side fix on the same commit kills the orphaned session at the other end too.

extensions/containers-canvas/src/webview/canvasVsCodeApi.js
```diff
@@ -0,0 +1,94 @@
...
+        window.postMessage(frame, "*");
+    });
+
+    return {
+        postMessage(message) {
+            void fetch(panelHref("./rpc"), {
+                method: "POST",
+                headers: { "Content-Type": "application/json" },
+                body: JSON.stringify(message),
+            });
```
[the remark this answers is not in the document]
patverb: Fixed in 7b3c0dd0 — a `.catch` is attached. Your framing is the reason it matters: this shim exists to notice the loopback server going away, and that is precisely the condition that makes the POST reject, so the failure it was written to watch for was surfacing as an unhandled rejection.

The rejection is swallowed deliberately rather than reported here, because the SSE stream's own error handling already drives the connection state; a comment records that so it does not read as an oversight later.

extensions/containers-canvas/src/webview/main.jsx
```diff
@@ -0,0 +1,1301 @@
...
+        </div>
+    );
+}
+
+function opsFor(item) {
+    if (item.kind !== "container") return [];
+    if (item.state === "paused") return ["stop", "restart", "unpause"];
+    if (item.state === "running") return ["stop", "restart", "pause"];
+    return ["start"];
+}
```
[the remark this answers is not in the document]
patverb: This one I think is incorrect, so flagging rather than changing anything. `opsFor` does include `remove` at the commit reviewed (a608d30a):

```js
function opsFor(item) {
    if (item.kind !== "container") return [];
    if (item.state === "paused") return ["stop", "restart", "unpause", "remove"];
    if (item.state === "running") return ["stop", "restart", "pause", "remove"];
    return ["start", "remove"];
}
```

Removal is reachable from the detail pane in every state, behind a confirmation dialog that names the container and shows the exact command — including the `-f` a running container needs. `verify:panel` asserts all of it against the daemon rather than the UI: that the verb is offered, that confirming destroys the container, and that the confirmation displayed `docker rm -f` when forcing.

The underlying point about images is fair, though: images have no removal control in the panel, only through the agent. Leaving that as a deliberate gap for now rather than widening this PR.

Resolving as already-addressed — happy to reopen if I have misread which revision this was against.

extensions/containers-canvas/src/webview/main.jsx
```diff
@@ -0,0 +1,1301 @@
...
+/**
+ * Placeholder shown while a sub-view's chunk is fetched.
+ *
+ * Deliberately quiet: the chunks are local and small, so on any normal machine
+ * this is a single frame. A prominent spinner would flash more than it informs.
+ */
+function ViewLoading() {
+    return (
+        <div style={{ padding: "24px", display: "flex", justifyContent: "center" }}>
+            <Spinner size="tiny" label="Loading…" />
```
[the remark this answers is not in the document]
patverb: Leaving this one open deliberately, because the gap is real even though the suggested remedy cannot be applied as written.

`vscode.l10n.t(...)` is not reachable from this package. A Copilot canvas runs outside VS Code, and the build aliases the bare `vscode` specifier to `src/vscode-stub.mjs`, whose members throw `'vscode.<name>' is not available outside VS Code`. So the concrete suggestion would fail at runtime rather than localize anything.

Your fallback is the actual ask — "a shared localization adapter for the non-VS-Code webview" — and I agree that is what it would take. It is also a good deal larger than the line flagged here: it means a webview-side message catalogue plus an extraction pipeline, applied to every label, error, placeholder and notification in the canvas, and a decision about how those bundles reach a plugin that installs straight from a git ref with no build step. The repository convention is real (137 files under `extensions/vscode-containers` use `l10n.t`), so this should not just be dropped.

Tracking it as follow-up rather than folding it into this PR, and leaving the thread unresolved so it stays visible.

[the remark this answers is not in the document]
patverb: This doesnt ship in vs. We can do that

bwateratmsft (commented):
We are probably gonna wanna turn on LFS for the big bundle files

extensions/containers-canvas/src/webview/main.jsx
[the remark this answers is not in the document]
patverb: Closing this. To be clear about what did and did not change: the code is unchanged, and the gap is real.

`vscode.l10n.t(...)` cannot be used here — a Copilot canvas runs outside VS Code, and this package aliases the bare `vscode` specifier to `src/vscode-stub.mjs`, whose members throw `'vscode.<name>' is not available outside VS Code`. So the specific remedy would fail at runtime rather than localize anything.

Your fallback is the real ask, and I agree with it: a shared localization adapter for the non-VS-Code webview. That is a catalogue plus an extraction step applied to every label, error, placeholder and notification in the canvas, and it needs a decision about how those bundles reach a plugin that installs straight from a git ref with no build step. Too large to fold into this pull request, and not something to do badly in a hurry.

Rather than leave it recorded only in a review thread, it is now written into the changelog's known limitations in 07898f26, where someone evaluating the plugin will actually see it.

extensions/containers-canvas/execSessions.mjs
```diff
@@ -0,0 +1,188 @@
...
+export function createExecSessions({ runtimeBin }) {
+    const sessions = new Map();
+
+    return {
+        /**
+         * Start a shell. `onData` receives raw terminal output including ANSI
+         * escapes -- xterm.js renders them, so nothing is stripped here.
+         */
+        async start({ containerId, shell, cols = 80, rows = 24, onData, onExit }) {
+            if (sessions.size >= MAX_SESSIONS) {
```
bwateratmsft: **🐛 Issue: Concurrent starts bypass the session limit**

🤖 Concurrent WebSocket connections can pass our eight-session check before either awaited operation completes; let's reserve capacity before awaiting or recheck immediately before spawning so pending starts cannot exceed the limit.

bwateratmsft (commented):
Reviewed with PeerPilot.

extensions/containers-canvas/execSessions.mjs
```diff
@@ -0,0 +1,188 @@
...
+
+        kill(id) {
+            const session = sessions.get(id);
+            if (!session) return;
+            sessions.delete(id);
+            try { session.proc.kill(); } catch { /* already gone */ }
+        },
+
+        /** Every session dies with the panel. A stray shell is a leaked process. */
+        disposeAll() {
```
bwateratmsft: **🐛 Issue: Disposal does not stop pending session starts**

🤖 If our panel closes during shell detection, startup can spawn a PTY after disposal because `wss.close()` leaves existing sockets open; let's mark the manager disposed and reject pending starts before spawning.

extensions/containers-canvas/extension.mjs
```diff
@@ -0,0 +1,213 @@
...
+                    limit: { type: "integer", minimum: 1, maximum: 500, description: "Max rows per list. Defaults to 50." },
+                },
+                additionalProperties: false,
+            },
+            handler: async (ctx) => {
+                const instance = instanceFor(ctx.instanceId);
+                const state = instance.getState() ?? (await instance.refresh());
+                const limit = ctx.input?.limit ?? 50;
+                const kind = ctx.input?.kind ?? "all";
+                return {
```
bwateratmsft: **🐛 Issue: Preserve runtime failures in the list response**

🤖 We should return `state.error` and runtime availability: `loadState` substitutes empty arrays when Docker is unreachable or listing commands fail, so this response incorrectly presents failed reads to Copilot as an empty inventory.

extensions/containers-canvas/skills/containers-canvas/SKILL.md
```diff
@@ -0,0 +1,107 @@
...
+| its ports, labels, mounts, env, and actions | `details` |
+| a general overview of everything | `list` |
+
+`view` defaults to `details` when a `target` is given and `list` otherwise. Pass
+`tab` (`containers` or `images`) to choose which list opens first.
+
+Examples:
+
+- "why did web stop?" → `{ target: "web", view: "logs" }`
+- "what's eating my memory?" → `{ view: "stats" }`, or add `target` for one
```
bwateratmsft: **🐛 Issue: Stats navigation requires a container target**

🤖 We should select a container and pass its `target` here: `extension.mjs` ignores targetless `view: "stats"`, leaving the list or previous pane visible rather than showing resource usage.

extensions/containers-canvas/src/webview/CodeView.jsx
```diff
@@ -0,0 +1,261 @@
...
+    const text = String(value ?? "");
+
+    const parsed = useMemo(() => {
+        if (language !== "json") return null;
+        const trimmed = text.trim();
+        if (!trimmed.startsWith("{") && !trimmed.startsWith("[")) return null;
+        try {
+            const v = JSON.parse(trimmed);
+            // `docker inspect` always returns an array; unwrap the single entry.
+            if (Array.isArray(v) && v.length === 1 && v[0] && typeof v[0] === "object") return v[0];
```
bwateratmsft: **🐛 Issue: Preserve JSON roots in file previews**

🤖 Let's preserve the root type in file previews and make inspect unwrapping opt-in: FilesView also uses this viewer, so `[{"enabled":true}]` appears as an object, while `[]` and `{}` both render blank.

extensions/containers-canvas/src/webview/ExecHistory.jsx
```diff
@@ -0,0 +1,162 @@
...
+ * Split a typed command into argv.
+ *
+ * Quotes are honoured so `echo "a b"` is two arguments, but nothing else is:
+ * there is no shell here, so `|`, `>` and `;` are ordinary characters. The
+ * placeholder says so, because silently treating them as literals would
+ * surprise anyone who types a pipe expecting it to work.
+ */
+export function splitArgv(text) {
+    const out = [];
+    const re = /"([^"]*)"|'([^']*)'|(\S+)/g;
```
bwateratmsft: **🐛 Issue: Quoted argument segments are split incorrectly**

🤖 We split `echo --label="hello world"` into three arguments instead of two, retaining literal quotes; let's tokenize quoted and unquoted segments together so embedded quotes preserve argument boundaries.

extensions/containers-canvas/src/webview/ExecHistory.jsx
```diff
@@ -0,0 +1,162 @@
...
+    }, [client, command, item.id, onRan]);
+
+    return (
+        <div>
+            <div className={styles.bar}>
+                <Input
+                    className={styles.grow}
+                    value={command}
+                    onChange={(_, d) => setCommand(d.value)}
+                    onKeyDown={(e) => { if (e.key === "Enter") run(); }}
```
bwateratmsft: **🐛 Issue: Enter bypasses the execution busy state**

🤖 We disable Run while executing but leave Enter active, allowing repeated keypresses to launch duplicate commands; let's guard in-flight execution in `run` and disable the input while busy.

extensions/containers-canvas/src/webview/FilesView.jsx
```diff
@@ -0,0 +1,334 @@
...
+    }, [client, item.id]);
+
+    useEffect(() => { list("/"); }, [list]);
+
+    const openFile = useCallback(async (filePath) => {
+        setReading(true);
+        setFileError(null);
+        setFile(null);
+        try {
+            setFile(await client.readFile.query({ id: item.id, path: filePath }));
```
bwateratmsft: **🐛 Issue: Prevent stale requests from replacing the current selection**

🤖 We leave rows clickable during requests, so an older read or listing can finish last and overwrite our latest selection; let's guard both handlers' result, error, and loading updates with request IDs.

extensions/containers-canvas/src/webview/DockerfileView.jsx
```diff
@@ -0,0 +1,214 @@
...
+    const colon = rest.indexOf(":");
+    return colon === -1
+        ? { repo, commit: rest, subdir: null }
+        : { repo, commit: rest.slice(0, colon), subdir: rest.slice(colon + 1) };
+}
+
+/** A browsable URL for the recorded source, when the host is one we can link. */
+function browseUrl(uri, entryPoint) {
+    const { repo, commit, subdir } = splitSourceUri(uri);
+    const gh = /^https?:\/\/github\.com\/([^/]+\/[^/.]+)/.exec(String(repo ?? ""));
```
bwateratmsft: **🐛 Issue: Preserve dots in GitHub repository names**

🤖 We truncate `https://github.com/org/service.api#abc` to `org/service`, linking to the wrong repository; let's capture the full repository segment and strip only a trailing `.git` suffix.

extensions/containers-canvas/src/webview/DockerfileView.jsx
```diff
@@ -0,0 +1,214 @@
...
+        try {
+            await navigator.clipboard.writeText(data?.reconstructed?.text ?? "");
+            onNotify?.("success", "Reconstruction copied.");
+        } catch {
+            onNotify?.("error", "Could not copy.");
+        }
+    };
+
+    const source = data?.provenance?.source ?? null;
+    const parts = source ? splitSourceUri(source.uri) : null;
```
bwateratmsft: **🐛 Issue: Display the separately recorded revision**

🤖 We ignore `source.revision`, which `imageProvenance` supplies from OCI revision labels and SLSA digests, hiding known commits when the URI has no fragment; let's use that revision as the commit fallback.

extensions/containers-canvas/src/webview/LogView.jsx
```diff
@@ -0,0 +1,346 @@
...
+    const [asking, setAsking] = useState(false);
+    const viewerRef = useRef(null);
+    const atBottomRef = useRef(true);
+
+    const load = useCallback(async () => {
+        setBusy(true);
+        setError(null);
+        try {
+            const res = await client.logs.query({ id: item.id, tail: Number(tail) });
+            setText(res.output ?? "");
```
bwateratmsft: **🐛 Issue: Discard obsolete one-shot log responses**

🤖 We can overwrite live output with an older snapshot when a pending refresh resolves after Follow starts; let's invalidate outstanding reads when following, the container, or the tail selection changes.

extensions/containers-canvas/src/webview/LogView.jsx
```diff
@@ -0,0 +1,346 @@
...
+     * The stream replays `tail` lines itself, so entering follow mode replaces
+     * the buffer rather than appending to what the one-shot read already
+     * fetched -- otherwise the last screenful appears twice.
+     *
+     * Chunks are not line-aligned: a write can split mid-line, so a partial
+     * trailing line is held back until its newline arrives.
+     */
+    useEffect(() => {
+        if (!following) return undefined;
+        setStreamError(null);
```
bwateratmsft: **🐛 Issue: Clear snapshot errors when following starts**

🤖 We retain a failed snapshot's error when Follow starts, so the viewer keeps displaying that error instead of successfully streamed logs; let's clear `error` when entering follow mode.

extensions/containers-canvas/src/webview/TerminalView.jsx
```diff
@@ -0,0 +1,224 @@
...
+            let frame;
+            try { frame = JSON.parse(event.data); } catch { return; }
+            if (frame.type === "data") term.write(frame.data);
+            else if (frame.type === "ready") { setStatus("connected"); setShell(frame.shell); setRisks(frame.risks ?? []); }
+            else if (frame.type === "error") { setError(frame.message); setStatus("closed"); }
+            else if (frame.type === "exit") {
+                setStatus("closed");
+                term.write(`\r\n\u001b[90m[session ended${frame.code ? ` — exit ${frame.code}` : ""}]\u001b[0m\r\n`);
+            }
+        });
```
bwateratmsft: **🐛 Issue: Ignore events from replaced sockets**

🤖 We should guard every socket callback against socketRef.current before processing events; otherwise a delayed close, error, or exit from a replaced connection can overwrite the new session's status and hide Disconnect while its PTY remains active.

extensions/containers-canvas/src/webview/TerminalView.jsx
```diff
@@ -0,0 +1,224 @@
...
+        socket.addEventListener("close", () => setStatus((s) => (s === "connected" ? "closed" : s)));
```
bwateratmsft: **🐛 Issue: End the connecting state when connection setup fails**

🤖 We leave the badge at connecting when the WebSocket handshake fails because error only sets a message and close only transitions connected; let's transition the current socket to closed on either failure.

extensions/containers-canvas/src/webview/canvasVsCodeApi.js
```diff
@@ -0,0 +1,100 @@
...
+
+    return {
+        postMessage(message) {
+            // Fire and forget, but not unhandled. This shim exists to notice the
+            // loopback server going away, and that is precisely when the POST
+            // rejects — so without a catch the failure it is watching for
+            // surfaces as an unhandled rejection instead. The SSE stream's own
+            // error handling is what reports the disconnection; here the
+            // rejection only needs to be swallowed deliberately.
+            void fetch(panelHref("./rpc"), {
```
bwateratmsft: **🐛 Issue: Wait for the response stream before sending RPCs**

🤖 We should queue RPC posts until SSE is open: during startup or reconnect, `/rpc` can execute before `stubPanel` has a sink, dropping the reply and leaving the calling UI waiting.

extensions/containers-canvas/src/webview/canvasVsCodeApi.js
```diff
@@ -0,0 +1,100 @@
...
+     * stream has been down long enough to mean the server is really gone.
+     */
+    let downSince = null;
+    source.addEventListener("open", () => {
+        downSince = null;
+        onControl({ type: "connection", ok: true });
+    });
+    source.addEventListener("error", () => {
+        if (source.readyState === EventSource.CLOSED || source.readyState === EventSource.CONNECTING) {
+            if (downSince === null) downSince = Date.now();
```
bwateratmsft: **🐛 Issue: Report terminal EventSource failures**

🤖 We should report `CLOSED` immediately or start a cancellable disconnect timer: a terminal failure such as `/events` returning 403 emits only one error, so this branch never shows our disconnected banner.

